PR #30, opened 2016-02-29
Adding podspec for GTMSessionFetcher

jayneflux:
Creating a pull request. Hopefully with the correct user ID vis-a-vis CLA.

### DIFF
--- a/GTMSessionFetcher.podspec
+++ b/GTMSessionFetcher.podspec
@@ -1,0 +1,51 @@
+# This file specifies the Pod setup for GTMSessionFetcher. It enables developers 
+# to import GTMSessionFetcher via the CocoaPods dependency Manager.
+Pod::Spec.new do |s|
+  s.name        = 'GTMSessionFetcher'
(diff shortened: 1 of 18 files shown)


GTMSessionFetcher.podspec
```diff
@@ -0,0 +1,51 @@
...
+  s.summary     = 'Google Toolbox for Mac - Session Fetcher.'
+  s.description = <<-DESC
+
+  GTMSessionFetcher makes it easy for Cocoa applications 
+  to perform http operations. The fetcher is implemented 
+  as a wrapper on NSURLSession, so its behavior is asynchronous 
+  and uses operating-system settings on iOS and Mac OS X.
+  DESC
+  
+  s.ios.deployment_target = '6.0'
```
perotinus: I think NSURLSession is only available for iOS 7+.

thomasvl:
Can you also squash these into a single commit so we just get the final spec coming in at once.

PR #38, opened 2016-03-31
Update GTMSessionFetcher.podspec

vlad-gemini:
Maybe it would be a good idea to follow the naming convention for the other Google pods (like gtm-oauth2, gtm-http-fetcher).


thomasvl:
Those other specs aren't official ones, for example https://github.com/google/gtm-oauth2/blob/master/GTMOAuth2.podspec

PR #61, opened 2016-09-16
Default NSURLSession delegate methods to a background queue.

thomasvl:
This should be transparent to client code; the fetcher delegate
methods/callbacks were already being called back on a separately specified
queue. Internally the fetcher supported using a background queue, and data
access was thread-safe, so this switches over the default and updates some
unit tests.


thomasvl:
And dropping until I can sort out the test failure, wasn't seeing them local, so must be a race still in there.

PR #70, opened 2016-12-05
Create new ifdef for reference to UIApplication.

welshm:
- Allows GTMSessionFetcherIOS to link to main application and extension with GTM_BACKGROUND_TASK_FETCHING enabled
- Allows APPLICATION_EXTENSION_API_ONLY to be set and let background tasks occur in main application

### DIFF
--- a/Source/GTMSessionFetcher.m
+++ b/Source/GTMSessionFetcher.m
@@ -819,11 +819,11 @@ static GTMSessionFetcherTestBlock GTM_NULLABLE_TYPE gGlobalTestBlock;
 #endif  // GTM_DISABLE_FETCHER_TEST_BLOCK
 
 #if GTM_BACKGROUND_TASK_FETCHING
+  id<GTMUIApplicationProtocol> app = [[self class] fetcherUIApplication];
   // Background tasks seem to interfere with out-of-process uploads and downloads.
-  if (!self.skipBackgroundTask && !self.useBackgroundSession) {
+  if (app && !self.skipBackgroundTask && !self.useBackgroundSession) {
     // Tell UIApplication that we want to continue even when the app is in the
     // background.
-    id<GTMUIApplicationProtocol> app = [[self class] fetcherUIApplication];
 #if DEBUG
     NSString *bgTaskName = [NSString stringWithFormat:@"%@-%@",
                             [self class], fetchRequest.URL.host];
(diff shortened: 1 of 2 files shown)


welshm:
Signed CLA

welshm:
Signed CLA (with @ritual.co domain)

Source/GTMSessionFetcher.m
```diff
@@ -819,11 +819,12 @@ - (void)beginFetchMayDelay:(BOOL)mayDelay
 #endif  // GTM_DISABLE_FETCHER_TEST_BLOCK
 
 #if GTM_BACKGROUND_TASK_FETCHING
-  // Background tasks seem to interfere with out-of-process uploads and downloads.
-  if (!self.skipBackgroundTask && !self.useBackgroundSession) {
+  // Background tasks seem to interfere with out-of-process uploads and downloads. Background
```
thomasvl: The background fetch support isn't out of process, it is just about using a UIApplication background task to allow it to finish and fire callbacks.

welshm: I reverted this part. I'm not familiar enough with how GTM uses background sessions (as opposed to UIApplication background tasks) to really update the comment here if something else is out of date.

I tried to amend my git change and push, but it seems to be having issues. I'm not very familiar with GitHub so it might take me a bit to get it to work properly.

Source/GTMSessionFetcher.m
```diff
@@ -819,11 +819,12 @@ - (void)beginFetchMayDelay:(BOOL)mayDelay
...
+  // tasks can only be started if an app exists to support them.
+  id<GTMUIApplicationProtocol> app = [[self class] fetcherUIApplication];
+  if (app && !self.skipBackgroundTask && !self.useBackgroundSession) {
```
thomasvl: Since we've got two CPP gates, I'd say this shuffle to ensure the we get an "app" to use is still a good safety net.

thomasvl:
Re multiple comments/pushing, no worries - I can squash it into a single comment when finally doing the merge.

Source/GTMSessionFetcher.m
welshm: SGTM. Done.

Manny27nyc (commented):
Thanks

PR #78, opened 2017-03-29
Xcode 8.3 support/updates

thomasvl:


thomasvl:
Hm, given the travis issues today, I'm not sure if I trust the results, likely will have to submit this and see what it does next. Things pass locally (except for some occasional flake that I don't want to disable around).

thomasvl:
And more flakes, opening an issue in general, as dealing with the flake problems is going to be a lot more than just this PR.

PR #86, opened 2017-06-06
Fix Xcode 9 Implicit conversion error.

wkiefer:
• Fixes `Implicit conversion from nullable pointer 'NSMutableData * _Nullable' to non-nullable pointer type 'ObjectType _Nonnull' (aka 'id')`


wkiefer:
Thanks @thomasvl ! We use this via https://github.com/google/google-api-objectivec-client-for-rest - how often do you update the submodule dependency on this repo? (or should I send a PR for that?)

thomasvl:
Are you building via pods or using the project directly?  If pods, the dep doesn't matter.

Sending a PR for that one module is fine.

PR #93, opened 2017-09-11
Sets default callback queue for recreated upload fetchers.

### LINKED ISSUE #92: Delegate Callback Queue not configured for restored background upload sessions
If you create an upload fetcher with the `useBackgroundSession` property set to `YES` and then resume the app while the upload is still in progress, the reconstructed fetcher does not send messages to its configured `sendProgressBlock`.

It looks the like the issue is that the reconstruction process doesn't configure the `_delegateCallbackQueue` property of the fetcher (whereas creating a new fetcher configures it to `self.callbackQueue` in `-[GTMSessionFetcher beginFetchWithCompletionHandler:]`).

I'm not sure if this is a bug or an issue that I'm not fully understanding. Configuring the delegate queue in `+[GTMSessionUploadFetcher uploadFetchersForBackgroundSessions]` starts propagating the calls to `sendProgressBlock` - around https://github.com/google/gtm-session-fetcher/blob/master/Source/GTMSessionUploadFetcher.m#L301, add

```
[uploadFetcher setDelegateCallbackQueue: uploadFetcher.callbackQueue];
```

mwyman:
Fixes #92, where callbacks are not being called on upload fetchers recreated from background session metadata, because the GTMSessionUploadFetchers created does not have a delegateCallbackQueue.

### DIFF
--- a/Source/GTMSessionUploadFetcher.m
+++ b/Source/GTMSessionUploadFetcher.m
@@ -221,6 +221,7 @@ NSString *const kGTMSessionFetcherUploadLocationObtainedNotification =
   uploadFetcher.sessionUserInfo = metadata;
   uploadFetcher.useBackgroundSession = YES;
   uploadFetcher.currentOffset = currentOffset;
+  uploadFetcher.delegateCallbackQueue = uploadFetcher.callbackQueue;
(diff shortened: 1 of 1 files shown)


zizajj:
All...

PR #96, opened 2017-09-27
Update project to support Xcode 9 and bumps version (2 commits)

sergiocampama:


Source/UnitTests/GTMSessionFetcherUtilityTest.m
```diff
@@ -132,7 +132,12 @@ - (void)testGTMFetcherStandardVersionString {
   NSBundle *bundle = [NSBundle bundleForClass:[self class]];
   NSString *result = GTMFetcherStandardUserAgentString(bundle);
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunguarded-availability"
+// Disable unguarded availability warning as we can't use the @availability macro until we require
+// all clients to build with Xcode 9 or above.
   NSOperatingSystemVersion version = [NSProcessInfo processInfo].operatingSystemVersion;
```
thomasvl: Actually, is this line only needed in the OS_IPHONE case?  Does that solve the warning?

thomasvl: Nevermind, looks like the mac one still needs it at the end.

PR #97, opened 2017-09-28
Update GTMSessionFetcher project to Xcode 9

sergiocampama:
Resolves warnings in Xcode 9 for project settings and test code.
Tested that everything builds with Xcode 8.3.3


Source/GTMSessionFetcherCore.xcodeproj/project.pbxproj
```diff
@@ -1313,7 +1321,7 @@
 		4FEE142C1BB22328002ACBF8 /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
-				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
+				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "";
```
thomasvl: I'm not sure we want this, won't it prevent it from being signed in someone uses the project in their build?

sergiocampama: I've returned those values. PTAL

PR #104, opened 2018-01-09
Rename framework targets to "GTMSessionFetcher".

WilliamDenniss:
- Previously had OS-specific suffixes like GTMSessionFetcherOSX.
- Benefit of using the same name for all platforms is that cross-platform codes can use the same include statements (e.g. `#import <GTMSessionFetcher/GTMSessionFetcher.h>`).
– This also harmonizes the Framework names with the ones generated by CocoaPods.


thomasvl:
The test flaky on travis makes it hard to see test failures vs. build failures at first glance, but clicking through to the logs shows the macOS ones not building.

thomasvl:
In Source there also are directories with the toplevel .h and Info.plist for each framework, that likely needs some cleanup since it would be a single toplevel header?  (but the name might conflict with the stock GTMSessionFetcher.h that isn't a framework wide header.

WilliamDenniss:
Hmm yes it seems that the issue is that the framework umbrella header needs to match the name of the framework, which I missed. But it it can't be renamed to GTMSessionFetcher.h due to the name conflict.

So we'd need a new platform-agnostic name for the framework that isn't GTMSessionFetcher (I assume renaming the stock GTMSessionFetcher.h is out of the question). What's your preference for the framework name?

thomasvl:
Yea, we can't rename the existing header since that would break all sorts of things.  Does the header name have to match the framework name?  i.e. - are we talking about a name for the header and a names for the framework or just a name for the header?

WilliamDenniss:
It would appear that it [does](https://developer.apple.com/library/content/documentation/MacOSX/Conceptual/BPFrameworks/Tasks/IncludingFrameworks.html). "The master header file is the header file whose name matches the name of the framework".

If we want to enable cross-platform framework includes then I think we need a new platform-agnostic name for the framework that isn't GTMSessionFetcher.  How about `GTMSessionFetcherFramework`? Then the include would be `#import <GTMSessionFetcherFramework/GTMSessionFetcherFramework.h>`.  This would distinguish the framework from the class.

thomasvl:
Closing this out since this simple swap likely won't work out, and improving the support for things like Carthage, etc. will take some more thought.

PR #130, opened 2018-10-24
Fix crash for GTMSessionFetcherService: in case of NSURSession.delegate was swizzled

gelosi:
Well, basically, the topic.

I've met this problem when I've used [TrustKit](https://github.com/datatheorem/TrustKit), which wraps NSURLSessionDelegate with own implementation.

So, I've re-created a test case, which crashes without the fix.

Have a look please if it works for all occasions (at least, tests went fine with me).

Looking forward to your comments.

### DIFF
--- a/Source/GTMSessionFetcherService.m
+++ b/Source/GTMSessionFetcherService.m
@@ -374,17 +374,19 @@ NSString *const kGTMSessionFetcherServiceSessionKey
 }
 
 // Internal utility. Returns a fetcher's delegate if it's a dispatcher, or nil if the fetcher
-// is its own delegate and has no dispatcher.
+// is its either own delegate and has no dispatcher or swizzled alien.
 - (GTMSessionFetcherSessionDelegateDispatcher *)delegateDispatcherForFetcher:(GTMSessionFetcher *)fetcher {
   GTMSessionCheckNotSynchronized(self);
 
   NSURLSession *fetcherSession = fetcher.session;
   if (fetcherSession) {
     id<NSURLSessionDelegate> fetcherDelegate = fetcherSession.delegate;
-    BOOL hasDispatcher = (fetcherDelegate != nil && fetcherDelegate != fetcher);
-    if (hasDispatcher) {
+      BOOL maybeHasDispatcher = (fetcherDelegate != nil && fetcherDelegate != fetcher);
+      
       GTMSESSION_ASSERT_DEBUG([fetcherDelegate isKindOfClass:[GTMSessionFetcherSessionDelegateDispatcher class]],
                               @"Fetcher delegate class: %@", [fetcherDelegate class]);
+      
+    if (maybeHasDispatcher && [fetcherDelegate isKindOfClass:GTMSessionFetcherSessionDelegateDispatcher.class]) {
       return (GTMSessionFetcherSessionDelegateDispatcher *)fetcherDelegate;
     }
   }
--- a/Source/GTMSessionFetcherService.m
+++ b/Source/GTMSessionFetcherService.m
@@ -383,9 +383,6 @@ NSString *const kGTMSessionFetcherServiceSessionKey
     id<NSURLSessionDelegate> fetcherDelegate = fetcherSession.delegate;
     BOOL maybeHasDispatcher = (fetcherDelegate != nil && fetcherDelegate != fetcher);
       
-    GTMSESSION_ASSERT_DEBUG([fetcherDelegate isKindOfClass:[GTMSessionFetcherSessionDelegateDispatcher class]],
-                            @"Fetcher delegate class: %@", [fetcherDelegate class]);
-    
     if (maybeHasDispatcher && [fetcherDelegate isKindOfClass:GTMSessionFetcherSessionDelegateDispatcher.class]) {
       return (GTMSessionFetcherSessionDelegateDispatcher *)fetcherDelegate;
     }
--- a/Source/GTMSessionFetcherService.m
+++ b/Source/GTMSessionFetcherService.m
@@ -381,11 +381,11 @@ NSString *const kGTMSessionFetcherServiceSessionKey
   NSURLSession *fetcherSession = fetcher.session;
   if (fetcherSession) {
     id<NSURLSessionDelegate> fetcherDelegate = fetcherSession.delegate;
-      BOOL maybeHasDispatcher = (fetcherDelegate != nil && fetcherDelegate != fetcher);
+    BOOL maybeHasDispatcher = (fetcherDelegate != nil && fetcherDelegate != fetcher);
       
-      GTMSESSION_ASSERT_DEBUG([fetcherDelegate isKindOfClass:[GTMSessionFetcherSessionDelegateDispatcher class]],
-                              @"Fetcher delegate class: %@", [fetcherDelegate class]);
-      
+    GTMSESSION_ASSERT_DEBUG([fetcherDelegate isKindOfClass:[GTMSessionFetcherSessionDelegateDispatcher class]],
+                            @"Fetcher delegate class: %@", [fetcherDelegate class]);
+    
     if (maybeHasDispatcher && [fetcherDelegate isKindOfClass:GTMSessionFetcherSessionDelegateDispatcher.class]) {
       return (GTMSessionFetcherSessionDelegateDispatcher *)fetcherDelegate;
     }
--- a/Source/UnitTests/GTMSessionFetcherURLSessionDelegateSwapTest.m
+++ b/Source/UnitTests/GTMSessionFetcherURLSessionDelegateSwapTest.m
@@ -110,6 +110,8 @@
     }];
     
     [self waitForExpectations:@[swizzleDidActuallyHappenExpectaion] timeout:1];
+    
+    [fetcher stopFetching];
 }
 
 @end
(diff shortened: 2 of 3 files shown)


gelosi:
@thomasvl Hi there :) I need help with testing the thing, if you have a moment please.
Test are failing sometimes – there's no explicit pre-requirements on env, and I'm not sure where to look for.
`testFetcherTestBlockSimulateDataCallbacks` - this guy has failed at my latest attempt

Source/GTMSessionFetcherService.m
```diff
@@ -374,17 +374,19 @@ - (void)startFetcher:(GTMSessionFetcher *)fetcher {
...
+// is its either own delegate and has no dispatcher or swizzled alien.
 - (GTMSessionFetcherSessionDelegateDispatcher *)delegateDispatcherForFetcher:(GTMSessionFetcher *)fetcher {
   GTMSessionCheckNotSynchronized(self);
 
   NSURLSession *fetcherSession = fetcher.session;
   if (fetcherSession) {
     id<NSURLSessionDelegate> fetcherDelegate = fetcherSession.delegate;
-    BOOL hasDispatcher = (fetcherDelegate != nil && fetcherDelegate != fetcher);
-    if (hasDispatcher) {
+      BOOL maybeHasDispatcher = (fetcherDelegate != nil && fetcherDelegate != fetcher);
```
thomasvl: Why the indent change here?  Did you mean for this to be part of some new condition?

Source/GTMSessionFetcherService.m
```diff
@@ -374,17 +374,19 @@ - (void)startFetcher:(GTMSessionFetcher *)fetcher {
...
+      
       GTMSESSION_ASSERT_DEBUG([fetcherDelegate isKindOfClass:[GTMSessionFetcherSessionDelegateDispatcher class]],
```
thomasvl: Why do the assert but let the code work when that isn't the case?  What are you trying to allow with the new test that you still want the assert also?

Source/UnitTests/GTMSessionFetcherURLSessionDelegateSwapTest.m
```diff
@@ -0,0 +1,117 @@
+//
```
thomasvl: Please use the standard header for sources, we can't take a patch with a specific copyright that doesn't follow the project licensing.

Source/GTMSessionFetcherService.m
```diff
@@ -374,17 +374,19 @@ - (void)startFetcher:(GTMSessionFetcher *)fetcher {
...
                               @"Fetcher delegate class: %@", [fetcherDelegate class]);
+      
+    if (maybeHasDispatcher && [fetcherDelegate isKindOfClass:GTMSessionFetcherSessionDelegateDispatcher.class]) {
```
thomasvl: Calling it out specifically - swizzling doesn't change the type, it changes the methods on a class, so I'm not following why someone swizzling was causing problems in the existing code and this is needed instead.

Source/GTMSessionFetcherService.m
gelosi: Ouch. Getting aligned with 2 space idents! Thanks.

Source/GTMSessionFetcherService.m
gelosi: True. I've kept this assert to bring focus to this issue. With the new code this assert should be replaced with the warning. Changing that.

Source/UnitTests/GTMSessionFetcherURLSessionDelegateSwapTest.m
gelosi: thanks

Source/GTMSessionFetcherService.m
gelosi: Thanks for finding a time for the MR review.
So, this is the actual problem pinpoint. 

My experiments with `TSKNSURLSessionDelegateProxy` revealed it's pretty well made wrapper for url session delegate.

What is does, in case you creating own session instance – it's wrapping your delegate into that proxy object, and then uses `forwardingTargetForSelector:` for unknown selectors.

Should work by idea. 
But the real thing happens in `fetcherDidBeginFetching:`:
```
// PROBLEM STARTS HERE: this class is not dispatcher, it's TrustKit wrapper, and, maybe 
// subwrapped if you lucky who's using facebook's sdk as well... whatever
// the reason why it's crashing → GTMSessionFetcher expects there to be either parent 
// GTMSessionFetcher (who's sharing it's session) or GTMSessionFetcherSessionDelegateDispatcher

    GTMSessionFetcherSessionDelegateDispatcher *delegateDispatcher =
          [self delegateDispatcherForFetcher:fetcher];

// Sanity check that the fetcher's session is the delegate's shared session.
    NSURLSession *sharedSession = delegateDispatcher.session;
    NSURLSession *fetcherSession = fetcher.session;

    // so, we're asked for NSURLSession → and we got it, but clean from any GTM*** code! 
    if (sharedSession != nil && fetcherSession == sharedSession) {
       /* some code here */
       NSURLSessionTask *task = fetcher.sessionTask;
      if (task) {
        // CRASH AT THE NEXT LINE:
        [delegateDispatcher setFetcher:fetcher
                               forTask:task];
      }
    }
```

because of `setFetcher:forTask:` call is sent now to the instance of TrustKit swizzled object, which will try to forward it to the wrapped object, and will fail → Crash.

thomasvl: > because of setFetcher:forTask: call is sent now to the instance of TrustKit swizzled object, which will try to forward it to the wrapped object, and will fail → Crash.

I'm still failing to see how what you describe has any swizzling.  You seemed to have described using a proxy delegate that forwards, but that doesn't require any swizzling.

And if they are just providing a proxy that forwards, isn't the -isKindOfClass: going to stop us from connecting things up?  i.e. - the proxy would come back `false` for that so we wouldn't return the delegate to have things relayed through.

eyalevy:
@thomasvl @gelosi 
Hi,
I have the same issue with TrustKit. I wrote example app that reproduce it:  https://github.com/eyalevy/trustkit-firebase-sample.
When I add this PR it fix this issue.
So when can I expect this PR to be merged?
Thanks

thomasvl:
> So when can I expect this PR to be merged?

There are still open questions about the change, as the descriptions seem to imply it isn't really swizzling, but something hooking out the delegate with a proxy. But exactly why it fails isn't clear either since if it was forwarding things to the real class they would work, so it doesn't seem right to accept this patch that could be disabling the calls to the delegate without a proper understanding of why they were failing in the first place.

gelosi:
@thomasvl agree. I need to build a nice proof on this change. Please, don't kill MR yet.

gelosi:
@thomasvl and yes, it will take it's time. But issue _is there_ 'cause this code does not check the class of the dispatcher delegate, assuming no one else can be a delegate of URLSession [especially, set itself as a delegate trough init method swizzling, as done by TrustKit for instance]

mwyman:
I'm in agreement with @thomasvl that this PR has not made clear what it's fixing and why disabling delegate calls is a proper solution. If TrustKit expects to replace internal behavior of other code (via swizzling or whatever), it's their responsibility to properly proxy the delegate, which it sounds like isn't happening; I'm not sure it's reasonable for every networking library to need to guard against that.

Source/GTMSessionFetcherService.m
bryce-newrelic: @thomasvl This ticket seems a little stale, but I've run into this issue as well with the New Relic iOS framework. We also swizzle NSURLSession and override the delegate object to inject a proxy. If I'm understanding this method correctly, `BOOL hasDispatcher = (fetcherDelegate != nil && fetcherDelegate != fetcher);`([GTMSessionFetcherService.m:384](https://github.com/google/gtm-session-fetcher/blob/master/Source/GTMSessionFetcherService.m#L384)) is asserting if `fetchDelegate` is not the `fetcher`, then it must be a `GTMSessionFetcherSessionDelegateDispatcher`. However, when a framework which instruments NSURLSession is introduced, proxying the delegate, this assumption is invalid. This results in the method `-setFetcher:forTask` potentially being called on a `GTMSessionFetcher` object. 

This is the issue I am currently experiencing. The proxied object is, in actuality, the fetcher, but because the check only compares the memory address, the memory address of `fetcherDelegate` will never match the memory address of `fetcher`. This results in the proxy object wrapping `fetcher` to be treated like a `GTMSessionFetcherSessionDelegateDispatcher` object. This results in -setFetcher:forTask to be called on the `GTMSessionFetcher` object.

I don't have full context, but using an 'isKindOfClass' maybe a better solution than a direct address check, unless the concern is there may be multiple unique instances of the GTMSessionFetcher object, but, if this was the case the logic flow would result in the same problem. 

I can't speak for geliso, but our proxy object is smart enough to verify 'isKindOfClass' against the wrapped delegate, and forward methods to that proxied object. So doing a 'isKindOfClass' check would perform as expected.

thomasvl: What you describe makes some sense.  But this PR should break things then, as for the setup you describe, the new `maybeHasDispatcher` would be `False` and we'd then hit the `return nil` case, which doesn't seem what would be desired.

So it sounds like the best bet might be to just do the class check hoping proxy code "gets it right", and that's it.

@mwyman does that sound reasonable to you?

gelosi: @thomasvl @mwyman looks like it is exactly desired behaviour in this case. So, swizzled delegate won't be used as gtm session internals.

mwyman: I think the class check in place of the pointer value check sounds reasonable.

thomasvl:
Can folks hitting this see if #142 fixes it for them?

thomasvl:
Hopefully #142 address the issues this was attempting to fix.

PR #138, opened 2019-01-22
check delegateDispatcher if it has setFetcher:forTask:

troublediehard:


troublediehard:
I signed it!

troublediehard:
@thomasvl can you help me here? Change is really simple I'm not sure if it is really breaking tests. Are there any instructions on how to run tests locally? I've tried simple run test targets but no luck.

thomasvl:
The test are generally flaky lately on travis, a bunch of the failures only seem to happen on VMs, so it is something timing related we haven't been able to dig into/sort out.

General comment on this PR, why is this needed?  The delegate isn't something that developers swap out.  So why wouldn't it have the selector implemented.  This seems a bit like https://github.com/google/gtm-session-fetcher/pull/130 where the change is because of something mucking with library internals, and there's no assurance the library will work if it can't reliably wire things together.

troublediehard:
Yes, it is my problem. I'm using TrustKit for certificate pinning in React Native project, so it is using swizzling. And it is crashing on the line mentioned https://github.com/google/gtm-session-fetcher/pull/130

```
      if (task) {
        // CRASH AT THE NEXT LINE:
        [delegateDispatcher setFetcher:fetcher
                               forTask:task];
      }
```

thomasvl:
In that case, then I have the same concern, since we're not wiring things up, it isn't clear to me how everything is always working out correctly.

troublediehard:
ok, make sense. I close this PR for now.

Issue #140, opened 2019-02-28
`GTMSessionUploadFetcher` endlessly sends empty chunks when `uploadFileURL` points to an empty or non-existing file


maksymmalyhin:
Issues like [this](https://github.com/firebase/firebase-ios-sdk/issues/2350) is an implication of this issue.

thomasvl:
What is _#2350_?  This project doesn't have an issue with that number.

maksymmalyhin:
@thomasvl Ah, sorry, I need to add a global link, since it is an issue from another project: https://github.com/firebase/firebase-ios-sdk/issues/2350

maksymmalyhin:
@thomasvl I think, I can provide a fix for the issue. Would you mind me doing that?

thomasvl:
> @thomasvl I think, I can provide a fix for the issue. Would you mind me doing that?

Nope, go for it!

thomasvl:
I believe that PR fixed this.

PR #141, opened 2019-03-01
GTMSessionUploadFetcher - Handle empty and non-existing files to upload (issue #140)

maksymmalyhin:
This is a fix for #140:
- empty files: return a server result (error or success depending on if the server accepts empty HTTP body for POST requests)
- unreachable file URL: complete with error


maksymmalyhin:
@thomasvl Here is how I see the fix of the issue #140. I Can see some tests are failing, I tried them locally in master, and they seem to fail there as well.

mwyman:
Can you sync this with google:master? That should deal with the test failures.

maksymmalyhin:
@mwyman Thank you, tests pass after merging master.

PR #142, opened 2019-05-13
Safely handle when the NSURLSession has been swizzled and its delegate replaced with a proxy.

mwyman:
Several third-party libraries are known to do this, which caused the delegateDispatcherForFetcher: to return an invalid dispatcher in certain cases, as the previous pointer-equality check didn't recognize the current session delegate as a proxy for the session fetcher itself.


bryce-newrelic:
This looks good to me 👍

PR #150, opened 2019-09-30
Reconnect background session fetchers after the {UI,NS}ApplicationDidFinishLaunching notification (iOS & Mac).

mwyman:
Background sessions have been reconnected in +[GTMSessionFetcher load], but that leads to loading from NSUserDefaults before the app delegate gets much chance to run (which is the starting point for most app code).

This change introduces GTMSESSION_RECONNECT_BACKGROUND_SESSIONS_ON_LAUNCH, which when defined to 1 (either explicitly or via the default behavior for iOS 13+ and macOS 10.15+) will reconnect after the app has launched, via registering for an applicationDidFinishLaunching notification inside +load. Apps can force the old behavior on iOS 13+ and macOS 10.15+ by defining this macro to 0.


Source/GTMSessionFetcher.m
```diff
@@ -212,9 +212,32 @@ @implementation GTMSessionFetcher {
...
+  NSNotificationCenter *nc = [NSNotificationCenter defaultCenter];
+  [nc addObserver:self
+         selector:@selector(reconnectFetchersForBackgroundSessionsOnAppLaunch:)
+             name:UIApplicationDidFinishLaunchingNotification
+           object:nil];
+#elif TARGET_OS_OSX
+  NSNotificationCenter *nc = [NSNotificationCenter defaultCenter];
+  [nc addObserver:self
+         selector:@selector(reconnectFetchersForBackgroundSessionsOnAppLaunch:)
+             name:NSApplicationDidFinishLaunchingNotification
```
thomasvl: Need a guarded `#import` of AppKit in this file (since the header doesn't need types from it (like it does with UIKit))

Source/GTMSessionFetcher.m
```diff
@@ -87,6 +90,34 @@
...
+//
+// The fetcher will automatically attempt to recreate the sessions on app
+// start, but doing so reads from NSUserDefaults. This may have launch-time
+// performance impacts.
+//
+// To avoid launch performance impacts, on iPhone/iPad/Mac with iOS 13+ and
+// macOS 10.15+ the GTMSessionFetcher class will register for the app launch
+// notification and perform the reconnect then.
+//
+// Apps targetting older SDKs can opt into the new behavior by defining
```
stephanemoore: "targeting"

PR #151, opened 2019-10-01
Move travis to Xcode 11.1

thomasvl:


thomasvl:
Since we set low min versions, things now work with newer xcodes, but we still need to fix things to support higher minversions.  Anyway, this now don't fail to compile, so it should be safe to land.

PR #152, opened 2019-10-02
Remove strong AppKit dependency introduced by PR #150

mwyman:
Removes the strong AppKit dependency by weak-linking the NSApplicationDidFinishLaunchingNotification symbol, rather than #import-ing AppKit.

This will allow linking the fetcher to command-line programs without requiring linking AppKit. I'm not sure it makes any sense for a non-AppKit linking app to be making out-of-process requests.

### DIFF
--- a/Source/GTMSessionFetcher.m
+++ b/Source/GTMSessionFetcher.m
@@ -18,9 +18,6 @@
 #endif
 
 #import "GTMSessionFetcher.h"
-#if TARGET_OS_OSX
-#import <AppKit/AppKit.h>
-#endif
 
 #import <sys/utsname.h>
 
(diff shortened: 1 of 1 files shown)


Source/GTMSessionFetcher.m
```diff
@@ -250,11 +247,16 @@ + (void)load {
...
+  // Weakly link the application-did-finish-launching notification name, since command-line
+  // Mac clients may not link AppKit.
+  extern NSNotificationName NSApplicationDidFinishLaunchingNotification __attribute__((weak_import));
+  if (NSApplicationDidFinishLaunchingNotification != NULL) {
+    NSNotificationCenter *nc = [NSNotificationCenter defaultCenter];
+    [nc addObserver:self
+           selector:@selector(reconnectFetchersForBackgroundSessionsOnAppLaunch:)
+               name:NSApplicationDidFinishLaunchingNotification
+             object:nil];
+  }
```
thomasvl: In theory, i guess we could do the `else` here and keep the old behavior, but I guess for commandline tools it doesn't really matter.

thomasvl: I believe the fail case here is while it made the import weak, it AppKit still needed to be passed for linking commandline tools, so while it made the runtime not needed, it was still build breaking in that one still had to pass AppKit to link.

stephanemoore: Ah, I believe you are correct. I suppose that the right approach would be to use `dlsym` to load the symbol?

PR #153, opened 2019-10-03
Revert the delayed resuming of background fetchers

thomasvl:
Still seeing some reports of issues with command line tools that used the fetcher, so rolling this back so it can be revisited to make things safer.


thomasvl:
@stephanemoore fyi

PR #154, opened 2019-10-07
Reconnect background session fetchers after UIApplicationDidFinishLaunching

mwyman:
Default on iOS 13+ to reconnecting background fetchers after the app launches, rather than directly +load.

Unlike the previous attempt (pr #150), this new behavior is only the default on iOS. Mac users must still explicitly enable the behavior, which requires linking the application with AppKit. If GTMSESSION_RECONNECT_BACKGROUND_SESSIONS_ON_LAUNCH=1 is defined for Mac, any background fetchers will be reconnected after NSApplicationDidFinishLaunching.

### DIFF
--- a/Source/GTMSessionFetcher.m
+++ b/Source/GTMSessionFetcher.m
@@ -92,32 +92,6 @@ GTM_ASSUME_NONNULL_END
   #endif
 #endif
 
-// When creating background sessions to perform out-of-process uploads and
-// downloads, on app launch any background sessions must be reconnected in
-// order to receive events that occurred while the app was not running.
-//
-// The fetcher will automatically attempt to recreate the sessions on app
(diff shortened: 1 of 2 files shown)


Source/GTMSessionFetcher.m
```diff
@@ -18,6 +18,11 @@
 #endif
 
 #import "GTMSessionFetcher.h"
+#if TARGET_OS_OSX && GTMSESSION_RECONNECT_BACKGROUND_SESSIONS_ON_LAUNCH
```
thomasvl: Actually, the `#define` is initialized below, so this isn't completely what you want.  Either it moves back to the .h or you likely need to move this down until after you default it.

mwyman: I moved the default define to the header file, since it's essentially part of the public API.

PR #161, opened 2019-10-15
Clean up macros guarding the use of TLSMinimumSupportedProtocol.

mwyman:
If the minimum deployment target is iOS 13+/watchOS 6+/tvOS 13+/macOS 10.15+ (which is required for Catalyst support, too), the @available() guard is not sufficient as the compiler no longer recognize the deprecated TLMinimumSupportedProtocol property.

Change the guard macro be two: a "requires" and a "supports" macro, and skip the @available() check if it's required.

### DIFF
--- a/Source/GTMSessionFetcher.m
+++ b/Source/GTMSessionFetcher.m
@@ -92,10 +92,21 @@ GTM_ASSUME_NONNULL_END
   #endif
 #endif
 
-#if ((TARGET_OS_IPHONE && defined(__IPHONE_13_0) && __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_0) || \
-     (TARGET_OS_OSX && defined(MAC_OS_X_VERSION_10_15) && MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_X_VERSION_10_15))
+#if ((defined(TARGET_OS_MACCATALYST) && TARGET_OS_MACCATALYST) || \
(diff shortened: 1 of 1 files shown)


Source/GTMSessionFetcher.m
```diff
@@ -92,10 +92,21 @@
   #endif
 #endif
 
-#if ((TARGET_OS_IPHONE && defined(__IPHONE_13_0) && __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_0) || \
-     (TARGET_OS_OSX && defined(MAC_OS_X_VERSION_10_15) && MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_X_VERSION_10_15))
+#if ((defined(TARGET_OS_MACCATALYST) && TARGET_OS_MACCATALYST) || \
+     (TARGET_OS_OSX && defined(__MAC_10_15) && __MAC_OS_X_VERSION_MIN_REQUIRED >= __MAC_10_15) || \
+     (TARGET_OS_IOS && defined(__IPHONE_13_0) && __IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_13_0) || \
```
thomasvl: are there `IOS` constants that should be use for all these rather than the mix of `IPHONE` and `IOS` ones?  (just thinking about how watchOS sdks still defines the `MIN_REQUIRED` one, but pinned to something old.

mwyman: I don't see ones besides the __IPHONE ones.

Regarding the watchOS SDK defining the iPhone min version, that only happens when the watch is targeted (they must need the iPhone min defined for that), so because we first check the targeted OS it should be fine.

PR #163, opened 2019-10-17
Use WKWebView in GTMSessionFetcherLogViewController, rather than UIWebView

mwyman:
UIWebView has been deprecated for some time. Use WKWebView in recent SDK builds.


Source/GTMSessionFetcherLogViewController.m
```diff
@@ -234,10 +254,10 @@ - (instancetype)initWithURL:(NSURL *)htmlURL
 }
 
 - (void)loadView {
-  UIWebView *webView = [[UIWebView alloc] init];
+  GTMWebView *webView = [[GTMWebView alloc] init];
   webView.autoresizingMask = (UIViewAutoresizingFlexibleWidth
                               | UIViewAutoresizingFlexibleHeight);
-  webView.delegate = self;
+  webView.navigationDelegate = self;
```
thomasvl: Doesn't this need an `#if` to use `delegate` or `navigationDelegate`?

mwyman: Yup. Thanks for catching!

PR #168, opened 2019-12-05
Fix flakey -testFetchToFile/testUnsuccessfulFetchToFile.

mwyman:
The download progress block was explicitly not being retained when dispatching to the callback queue, meaning the callback may or may not be called, depending on the dispatch behavior. This led to the test flaking, because it would not always update the total bytes written.


thomasvl:
_retained_ here is not from a ARC pov, but more a _cached_ pov, i.e. things complete and the progress block gets cleared by the fetcher, so when the block finally went to do the callback, the block was gone.  So fetching in before hand is what ensures we can still call it for that _final_ notification.

Also, since this _final_ progress will fire more often, there is a minor behavior change, but it will be well behaved in that it happens before completion.

PR #173, opened 2019-12-18
Cleanup iOS 8/macOS 10.10 version checking cruft.

mwyman:
A bunch of now-unnecessary #ifs were added at one point due to @available not being universally available. Clients are now assumed to be building with a version of Xcode that has @available, so cleaned up the pile of #ifs and replaced with some @available checks.

No functional changes.

### DIFF
--- a/Source/GTMSessionFetcher.m
+++ b/Source/GTMSessionFetcher.m
@@ -639,36 +639,19 @@ static GTMSessionFetcherTestBlock GTM_NULLABLE_TYPE gGlobalTestBlock;
         NSMapTable *sessionIdentifierToFetcherMap = [[self class] sessionIdentifierToFetcherMap];
         [sessionIdentifierToFetcherMap setObject:self forKey:self.sessionIdentifier];
 
-#if (TARGET_OS_TV \
-     || TARGET_OS_WATCH \
-     || (!TARGET_OS_IPHONE && defined(MAC_OS_X_VERSION_10_10) && MAC_OS_X_VERSION_MIN_REQUIRED >= MAC_OS_X_VERSION_10_10) \
-     || (TARGET_OS_IPHONE && defined(__IPHONE_8_0) && __IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_8_0))
-        // iOS 8/10.10 builds require the new backgroundSessionConfiguration method name.
-        _configuration =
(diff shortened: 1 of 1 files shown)


Source/GTMSessionFetcher.m
```diff
@@ -639,36 +639,19 @@ - (void)beginFetchMayDelay:(BOOL)mayDelay
...
-    || (TARGET_OS_IPHONE && defined(__IPHONE_8_0) && __IPHONE_OS_VERSION_MIN_REQUIRED < __IPHONE_8_0)
-        // Do a runtime check to avoid a deprecation warning about using
-        // +backgroundSessionConfiguration: on iOS 8.
-        if ([NSURLSessionConfiguration respondsToSelector:@selector(backgroundSessionConfigurationWithIdentifier:)]) {
-          // Running on iOS 8+/OS X 10.10+.
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wunguarded-availability"
-// Disable unguarded availability warning as we can't use the @availability macro until we require
-// all clients to build with Xcode 9 or above.
+        if (@available(iOS 8.0, macOS 10.10, *)) {
```
thomasvl: While this implicity covers watchOS and tvOS, should we add them explicit so someone doesn't know how iOS also covers those other platforms?

Source/GTMSessionFetcher.m
```diff
@@ -639,36 +639,19 @@ - (void)beginFetchMayDelay:(BOOL)mayDelay
...
           _configuration =
               [NSURLSessionConfiguration backgroundSessionConfigurationWithIdentifier:sessionIdentifier];
-#pragma clang diagnostic pop
         } else {
-          // Running on iOS 7/OS X 10.9.
+#if ((!TARGET_OS_IPHONE && MAC_OS_X_VERSION_MIN_REQUIRED < MAC_OS_X_VERSION_10_10) \
```
thomasvl: Why is this gate needed?  To avoid deprecated warnings when the min is high enough the first check is always true?

Source/GTMSessionFetcher.m
mwyman: Alright.

Source/GTMSessionFetcher.m
mwyman: Yes. Even though then compiler will omit the else clause (and even the version check) if it knows the minimum target is high enough, it still warns/errors about using a deprecated method in the else case.

PR #181, opened 2020-01-29
Pass testBlock onto subordinate upload fetchers.

tfg3:
With testBlock specified, GTMSessionUploadFetcher's subordinate query and chunk fetchers attempt to hit the network. This change passes testBlock onto subordinates, allowing test callers to simulate various backend responses and network errors at query and chunk upload time.


mwyman:
After a discussion offline, we believe this PR is aimed to address #185, but is probably insufficient to solve that issue. I'm looking into options, possibly including a new test block type for the upload fetcher. Unfortunately this gets complicated because even with a test block, the block is allowed to call its response with all nils which should continue the fetch; also that the test block could use the global test block, which is not exposed through to the upload fetcher.

For now I'd suggest we close this PR and move the discussion over to #185.

Issue #80, opened 2017-03-30
UnitTest Flake / Timing issues


thomasvl:
Running the tests with Xcode 8.3 almost always passes, but under automation (travis or otherwise), we see a lot of flakes.  If the automation uses VMs (travis), it seems like they are even more common.  So odds are we've got timing issues within the tests.

Cheecherngho799:
Automation of flakes

Cheecherngho799:
Jailed

thomasvl:
#182 helps with this.

thomasvl:
Hopefully fixed with #183.

PR #182, opened 2020-01-30
Guard tests against race condition when asserting callbacks released.

mwyman:
Because the fetcher callbacks are released after the fetcher dispatches its final callbacks to the callback queue, our changing the default sessionDelegateQueue from the mainQueue to a background queue introduced a race condition to the tests: asserting the callbacks were nil when it was possible the session delegate queue had not yet had a chance to release them.

This adds a bounce through the session delegate queue before allowing the assert, to ensure the fetcher has had a chance to complete releasing the callbacks.


mwyman:
Good point, should get tagged for #80, this and the other PR.

PR #186, opened 2020-02-06
Fix background task behavior with background sessions.

mwyman:
Previously was checking `useBackgroundSession` for the active state, while that property is documented only as the client's expressed preference that the fetch use a background session. Even with `useBackgroundSession = NO`, if a `sessionIdentifer` is set before beginning the fetch than a background session will be used (and `usingBackgroundSession` will be `YES`. Because `useBackgroundSession` was being checked instead, a background task could be created but shouldn't have been.

This also fixes some issues with GTMSessionUploadFetcher instances and their chunk fetchers. For the initial fetch, `canFetchWithBackgroundSession` will return NO, but `useBackgroundSession` could be YES, leading to _not_ creating a background task identifier for the initial fetch, but because subsequent chunk fetchers are created without `useBackgroundSession == YES`, but relying on the `sessionIdentifier` being non-nil to use a background session, this would lead to creating background task IDs for the chunk fetchers.

Also fixed a potential issue if a fetch for a data: URL is created and `useBackgroundSession == YES`.

### DIFF
--- a/Source/GTMSessionFetcher.m
+++ b/Source/GTMSessionFetcher.m
@@ -551,8 +551,11 @@ static GTMSessionFetcherTestBlock GTM_NULLABLE_TYPE gGlobalTestBlock;
                            fetchRequest);
     }
 #endif
+    // If priorSessionIdentifier is allowed to stay nil, it will still create a background
+    // session.
+    priorSessionIdentifier = nil;
     [self setSessionIdentifierInternal:nil];
-    self.useBackgroundSession = NO;
+    self.usingBackgroundSession = NO;
   }
 
 #if GTM_ALLOW_INSECURE_REQUESTS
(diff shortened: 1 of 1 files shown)


Source/GTMSessionFetcher.m
```diff
@@ -551,8 +551,11 @@ - (void)beginFetchMayDelay:(BOOL)mayDelay
                            fetchRequest);
     }
 #endif
+    // If priorSessionIdentifier is allowed to stay non-nil, a background session can
```
thomasvl: Would the previous try not have been background if it was a data url?  And you can't change the url after the start, so what would have gotten it into this state?

mwyman: The concern is that priorSessionIdentifier is what's [checked below](https://github.com/google/gtm-session-fetcher/blob/96b3d35b30a983f152d4cc11db0adb45adf363eb/Source/GTMSessionFetcher.m#L634), so if a data URL is provided but a session identifier were to be set, we could still get in a situation where we attempt to create a background session if `useBackgroundSession` is true. Also, by changing `useBackgroundSession` and not `usingBackgroundSession` we're in the same boat.

Issue #187, opened 2020-03-10
Feature request: support for metrics collection


lukhnos:
The `NSURLSessionTaskDelegate` protocol has a new method [`-URLSession:task:didFinishCollectingMetrics:`](https://developer.apple.com/documentation/foundation/nsurlsessiontaskdelegate/1643148-urlsession?language=objc) since iOS 10.0, which is invoked when metrics have been collected for the session. Users who are interested in things like [connection time and transfer stats](https://developer.apple.com/documentation/foundation/nsurlsessiontasktransactionmetrics?language=objc) can then just rely on the information collected by the native network stack instead of rolling their own measurement routines.

It appears that one should make changes to [`GTMSessionFetcherSessionDelegateDispatcher`](https://github.com/google/gtm-session-fetcher/blob/master/Source/GTMSessionFetcherService.m#L46) as well as the downstream code path and introduce a new callback block type to the fetcher itself for those interested in getting the metrics information.

I have a proposed change and I'll make a pull request shortly.

lukhnos:
Resolved via #188

PR #188, opened 2020-03-10
Support metrics collection. (#187)

lukhnos:
This adds support to handle the
`-URLSession:task:didFinishCollectingMetrics:` delegate method in the
`NSURLSessionTaskDelegate protocol`. A new callback block type,
`GTMSessionFetcherMetricsCollectionBlock`, is also added.

This feature is enabled when compiled for iOS 10+, macOS 10.12+, Mac
Catalyst 13.0+, tvOS 10.0+, or watchOS 3.0+.

New test methods are added to cover this feature.

This commit also upgrades the iOS target SDK in the Xcode project file
to iOS 10.0.

### DIFF
--- a/Source/GTMSessionFetcher.h
+++ b/Source/GTMSessionFetcher.h
@@ -560,6 +560,9 @@ typedef void (^GTMSessionFetcherRetryBlock)(BOOL suggestedWillRetry,
                                             NSError * GTM_NULLABLE_TYPE error,
                                             GTMSessionFetcherRetryResponse response);
 
+API_AVAILABLE(ios(10.0), macosx(10.12), tvos(10.0), watchos(3.0))
+typedef void (^GTMSessionFetcherMetricsCollectionBlock)(NSURLSessionTaskMetrics *metrics);
+
 typedef void (^GTMSessionFetcherTestResponse)(NSHTTPURLResponse * GTM_NULLABLE_TYPE response,
                                               NSData * GTM_NULLABLE_TYPE data,
                                               NSError * GTM_NULLABLE_TYPE error);
@@ -995,6 +998,13 @@ NSData * GTM_NULLABLE_TYPE GTMDataFromInputStream(NSInputStream *inputStream, NS
 // fetch.
 // See comments at the top of this file.
 @property(atomic, copy, GTM_NULLABLE) GTMSessionFetcherRetryBlock retryBlock;
+
+// The optional block for collecting the metrics of the present session.
+//
+// This is called on the callback queue.
+@property(atomic, copy, GTM_NULLABLE)
+    GTMSessionFetcherMetricsCollectionBlock metricsCollectionBlock API_AVAILABLE(
+        ios(10.0), macosx(10.12), tvos(10.0), watchos(3.0));
 
 // Retry intervals must be strictly less than maxRetryInterval, else
 // they will be limited to maxRetryInterval and no further retries will
--- a/Source/GTMSessionFetcher.m
+++ b/Source/GTMSessionFetcher.m
@@ -1795,6 +1795,9 @@ NSData * GTM_NULLABLE_TYPE GTMDataFromInputStream(NSInputStream *inputStream, NS
   self.retryBlock = nil;
   self.testBlock = nil;
   self.resumeDataBlock = nil;
+  if (@available(iOS 10.0, macOS 10.12, tvOS 10.0, watchOS 3.0, *)) {
+    self.metricsCollectionBlock = nil;
+  }
 }
 
 - (void)forgetSessionIdentifierForFetcher {
@@ -2853,6 +2856,21 @@ didCompleteWithError:(NSError *)error {
   }];
 }
 
+- (void)URLSession:(NSURLSession *)session
+                          task:(NSURLSessionTask *)task
+    didFinishCollectingMetrics:(NSURLSessionTaskMetrics *)metrics
+    API_AVAILABLE(ios(10.0), macosx(10.12), tvos(10.0), watchos(3.0)) {
+  @synchronized(self) {
+    GTMSessionMonitorSynchronized(self);
+    GTMSessionFetcherMetricsCollectionBlock metricsCollectionBlock = _metricsCollectionBlock;
+    if (metricsCollectionBlock) {
+      [self invokeOnCallbackQueueUnlessStopped:^{
+        metricsCollectionBlock(metrics);
+      }];
+    }
+  }
+}
+
 #if TARGET_OS_IPHONE
 - (void)URLSessionDidFinishEventsForBackgroundURLSession:(NSURLSession *)session {
   GTM_LOG_SESSION_DELEGATE(@"%@ %p URLSessionDidFinishEventsForBackgroundURLSession:%@",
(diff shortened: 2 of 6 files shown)


lukhnos:
> I think the pbxproj file can be reverted in total; changing the deployment target to 10.0 is a separate discussion.

Done. Also wrapped the call to `-invokeOnCallbackQueueUnlessStopped:` with `@synchornized(self)`, otherwise an internal `GTMSessionCheckSynchronized()` check would complain.

Source/GTMSessionFetcher.m
```diff
@@ -2853,6 +2857,22 @@ - (void)URLSession:(NSURLSession *)session
   }];
 }
 
+#if GTM_SUPPORTS_METRICS_COLLECTION
+- (void)URLSession:(NSURLSession *)session
+              task:(NSURLSessionTask *)task
+didFinishCollectingMetrics:(NSURLSessionTaskMetrics *)metrics {
+  GTMSessionFetcherMetricsCollectionBlock metricsCollectionBlock = _metricsCollectionBlock;
```
mwyman: This should happen inside the @synchronized block.

Source/GTMSessionFetcher.h
```diff
@@ -441,6 +441,16 @@ extern "C" {
   #endif
 #endif
 
+#if ((defined(TARGET_OS_MACCATALYST) && TARGET_OS_MACCATALYST) || \
+     (TARGET_OS_OSX && defined(__MAC_10_12) && __MAC_OS_X_VERSION_MIN_REQUIRED >= __MAC_10_12) || \
+     (TARGET_OS_IOS && defined(__IPHONE_10_0) && __IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_10_0) || \
+     (TARGET_OS_WATCH && defined(__WATCHOS_3_0) && __WATCHOS_VERSION_MIN_REQUIRED >= __WATCHOS_3_0) || \
+     (TARGET_OS_TV && defined(__TVOS_10_0) && __TVOS_VERSION_MIN_REQUIRED >= __TVOS_10_0))
+#define GTM_SUPPORTS_METRICS_COLLECTION 1
```
mwyman: This is usually necessary only for new Xcode versions, but NSURLSessionTaskMetrics has been around since iOS 10, meaning Xcode 8, and we don't support building with such an old Xcode anymore. I think we can skip this, and if necessary simply annotate methods with API_AVAILABLE to catch any attempts to call on older OS versions.

I think it would really be fine to allow setting the metricsCollectionBlock property even on older versions of the OS, as long as the property/block is documented as only being called if the OS supports metrics collection.

lukhnos: > This is usually necessary only for new Xcode versions, but NSURLSessionTaskMetrics has been around since iOS 10, meaning Xcode 8, and we don't support building with such an old Xcode anymore.

I like the idea of not needing the `GTM_SUPPORTS_METRICS_COLLECTION` macro, but the problem is that `NSURLSessionTaskMetrics` is only available when building with iOS 10+ SDK, and the iOS targets in GTMSessionFetcher's Xcode project is still set to iOS 8. What would be the best practice to address this?

Source/GTMSessionFetcher.m
lukhnos: Done.

Source/GTMSessionFetcher.h
mwyman: The thing is, we should not have any clients at this time building with an SDK < iOS 12 (based on Apple's Xcode submission requirements), and fairly soon (IIRC) everyone should be building with the iOS 13 SDK. As a result, the symbols will exist at compile time. If we mark the property with API_AVAILABLE(ios(10.0)) etc, anyone targeting less than that would need to wrap the code in @available(). Because it's a class, we can compile with it and as long as it's not referenced from < iOS 10 code, it'll be fine.

We may need to similarly annotate one or more internal methods, but that shouldn't be problematic.

Source/GTMSessionFetcher.m
```diff
@@ -2853,6 +2857,22 @@ - (void)URLSession:(NSURLSession *)session
...
 }
 
+#if GTM_SUPPORTS_METRICS_COLLECTION
+- (void)URLSession:(NSURLSession *)session
+              task:(NSURLSessionTask *)task
+didFinishCollectingMetrics:(NSURLSessionTaskMetrics *)metrics {
+  @synchronized(self) {
+    GTMSessionFetcherMetricsCollectionBlock metricsCollectionBlock = _metricsCollectionBlock;
+    if (metricsCollectionBlock) {
+      GTMSessionMonitorSynchronized(self);
```
mwyman: Move this GTMSessionMonitorSynchronized to immediately after the @synchronized.

lukhnos: Done.

Source/GTMSessionFetcher.h
lukhnos: Done. I wasn't sure about the right way to add `API_AVAILABLE` to typedefs, but Xcode showed me how.

lukhnos:
I've addressed the comments so far, please take a look. Thanks!

Source/UnitTests/GTMSessionFetcherFetchingTest.m
```diff
@@ -1849,6 +1852,102 @@ - (void)testInsecureRequests_WithoutFetcherService {
   [self testInsecureRequests];
 }
 
+- (void)testCollectingMetrics_WithSuccessfulFetch API_AVAILABLE(ios(10.0), macosx(10.12),
```
mwyman: We generally have versions of all the tests both with and without a fetcher service (_WithoutFetcherService), so you should also have versions of these without (the default is with).

Source/GTMSessionFetcher.h
```diff
@@ -996,6 +999,13 @@ NSData * GTM_NULLABLE_TYPE GTMDataFromInputStream(NSInputStream *inputStream, NS
 // See comments at the top of this file.
 @property(atomic, copy, GTM_NULLABLE) GTMSessionFetcherRetryBlock retryBlock;
 
+// The optional block for collecting the metrics of the present session.
+//
+// This is called on the callback queue.
+@property(atomic, copy, GTM_NULLABLE)
```
mwyman: These blocks are generally available to set once on the fetcher service, and are propagated to each fetcher created by that service (the client code can always individually set a different one later). Should do that with this as well.

Source/UnitTests/GTMSessionFetcherFetchingTest.m
lukhnos: Done.

Source/GTMSessionFetcher.h
lukhnos: Done.

lukhnos:
The latest test failure appears to be due to flakiness (it passed locally) and did not involve the changes here. Could https://github.com/google/gtm-session-fetcher/pull/189 fix that already? At any rate, PTAL.

Source/GTMSessionFetcher.m
```diff
@@ -3462,6 +3480,7 @@ + (GTM_NULLABLE GTMSessionFetcherSystemCompletionHandler)systemCompletionHandler
             sendProgressBlock = _sendProgressBlock,
             willCacheURLResponseBlock = _willCacheURLResponseBlock,
             retryBlock = _retryBlock,
+            metricsCollectionBlock = _metricsCollectionBlock,
```
mwyman: Please set self.metricsCollectionBlock = nil inside releaseCallbacks.

lukhnos: It's already there in this PR: https://github.com/google/gtm-session-fetcher/pull/188/files#diff-590105fbe264a2024b0ac16eec191826R1799

lukhnos:
> Please sync with the master branch, as that may resolve these test failures (which seem unrelated to this PR). Otherwise, I think the other comment I have should be the only other change.

Rebased on the latest master. Thanks for the review!

PR #189, opened 2020-03-20
Fix upload fetcher when using upload data

mwyman:
Broken by PR #141.


mwyman:
Looks like a build timed out.

PR #191, opened 2020-04-29
Add Swift Package Manager support

karimhm:
~~To test changes in this PR using CocoaPods, please use this source: **https://github.com/karimhm/gtm-session-fetcher-specs**~~

~~Updated **GTMSessionFetcher.podspec.json** that is compatible with changes in this PR can be found [**here**](https://github.com/karimhm/gtm-session-fetcher-specs/blob/master/GTMSessionFetcher/1.4.0/GTMSessionFetcher.podspec.json)~~. This is no longer needed.


thomasvl:
Can you not change the source layout and instead use the Package to explicitly call out the sources?  We've always [documented](https://github.com/google/gtm-session-fetcher/wiki#adding-the-fetcher-to-your-project) what sources are needed for what features and there are people that integrate that way, so changing the layout on disk is a breaking change for those folks.

Source/UnitTests/GTMSessionFetcherFetchingTest.m
```diff
@@ -91,6 +91,40 @@ - (NSString *)docRootPath {
   NSBundle *testBundle = [NSBundle bundleForClass:[self class]];
   XCTAssertNotNil(testBundle);
 
+#if SWIFT_PACKAGE
+  // Swift Pacage Manager does not support resources distribution currently,
+  // therefore the "gettysburgaddress.txt" file must be copied manually to the test bundle.
+  static dispatch_once_t onceToken;
```
thomasvl, line 97: I guess this works because SwiftPM only runs test on macOS?  Writing into the test bundle would break the signature, so that won't work for other platforms.

Source/UnitTests/GTMSessionFetcherTestServer.m
```diff
@@ -220,6 +220,15 @@ - (NSURL *)localURLForFile:(NSString *)name {
   return [NSURL URLWithString:urlString];
 }
 
+- (NSURL *)localURLForFileUsingAppend:(NSString *)name {
+  // We need to create http URLs referring to the desired
+  // resource to be found by the http server running locally.
+  // Return a localhost:port URL for the test file
+
+  NSString *urlString = [NSString stringWithFormat:@"http://localhost:%d", _server.port];
+  return [[NSURL URLWithString:urlString] URLByAppendingPathComponent:name];
```
thomasvl, line 229: Isn't this the same results as calling the existing helper?

Source/UnitTests/GTMSessionFetcherUtilityTest.m
```diff
@@ -120,6 +120,10 @@ - (void)testGTMFetcherSystemVersionString {
 }
 
 - (void)testGTMFetcherApplicationIdentifier {
+// Swift Packages do not have the notion of bundles,
+// and the actual identifier might change from release to release,
+// or wether was built using Xcode or swift command.
```
thomasvl, line 125: How is the bundle id changing? it would be nice to still get test coverage for these things.

karimhm:
> Can you not change the source layout and instead use the Package to explicitly call out the sources? We've always [documented](https://github.com/google/gtm-session-fetcher/wiki#adding-the-fetcher-to-your-project) what sources are needed for what features and there are people that integrate that way, so changing the layout on disk is a breaking change for those folks.

We will need to add symbolic links in the new folders pointing to files outside of it. Swift PM works best when the code is structured into directories.

thomasvl:
Links seems sub optimal also, I thought swiftpm was supposed to have support to directly list files for supporting legacy layouts?

Source/UnitTests/GTMSessionFetcherTestServer.m
karimhm: When the tests are run using `swift test` command, trying to initialize an NSURL with a string that is not percent encoded will fail and the result will be nil.

Source/UnitTests/GTMSessionFetcherUtilityTest.m
karimhm: When running the tests using `swift test` command, the bundle identifier will be **proc_xctest**. And if the same test is run under Xcode it will be **GTMSessionFetcherCoreTests**. it might be possible to change the bundle by passing some compiler flags, but this approach is hacky and might break in the future.

Source/UnitTests/GTMSessionFetcherFetchingTest.m
karimhm: > Links seems sub optimal also, I thought swiftpm was supposed to have support to directly list files for supporting legacy layouts?

Swift PM will generate an error when 2 targets share the same **path**

Source/UnitTests/GTMSessionFetcherServiceTest.m
```diff
@@ -737,8 +737,11 @@ - (void)testUserAgentFromSessionConfiguration {
 
   XCTestExpectation *expectReceiveResponse = [self expectationWithDescription:@"Received response"];
   NSString *gettysburgPath = [_testServer localPathForFile:kValidFileName];
-  NSString *echoHeadersPath = [gettysburgPath stringByAppendingString:@"?echo-headers=true"];
-  NSURL *validFileURL = [_testServer localURLForFile:echoHeadersPath];
+  NSURL *gettysburgFileURL = [_testServer localURLForFileUsingAppend:gettysburgPath];
+  NSURLComponents *validFileURLComponents = [[NSURLComponents alloc] initWithURL:gettysburgFileURL
+                                                         resolvingAgainstBaseURL:NO];
+  validFileURLComponents.query = @"?echo-headers=true";
+  NSURL *validFileURL = validFileURLComponents.URL;
```
karimhm, line 744: The NSURL [query](https://developer.apple.com/documentation/foundation/nsurl/1407543-query) property will be nil if the parameters are appended directly, which will result is a failed test.

Source/UnitTests/GTMSessionFetcherFetchingTest.m
karimhm: Swift PM does not support running tests on real devices. I got this error when tried to run them in my device

```
Testing couldn’t begin on the device “iPhone”: Logic Testing Unavailable
Logic Testing on iOS devices is not supported. You can run logic tests on the Simulator.
```

karimhm:
> Can you not change the source layout and instead use the Package to explicitly call out the sources? We've always [documented](https://github.com/google/gtm-session-fetcher/wiki#adding-the-fetcher-to-your-project) what sources are needed for what features and there are people that integrate that way, so changing the layout on disk is a breaking change for those folks.

The layout was restored to its original state.

Source/UnitTests/GTMSessionFetcherServiceTest.m
thomasvl: hm, odd, but ok.

.travis.yml
```diff
@@ -1,19 +1,29 @@
...
+      # Build the project using Swift PM
+        - swift build
+      # Test the project using Swift PM
+        - swift test
 
-branches:
-  except:
-    - google
+    - stage: Xcode Tests
+      env:
```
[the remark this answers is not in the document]
karimhm: What about updating **.travis.sh** to be like: `{iOS|OSX|tvOS|SwiftPM} {Debug|Release|Both}`?, and then the Swift tests will be added to the rest of the test suite.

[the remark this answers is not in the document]
thomasvl: That can also work, but the logic in the rest of the script will be a bit more complex because it can't simply waterfall thru to building up an `xcodebuild` command line; but something like that should be possible.

If you want to just do the simple travis.yml change for now, we can look at updating the script later to get debug and release coverage for SwiftPM also.

karimhm:
@thomasvl  A new release is needed to allow successful Swift PM checkouts, e.g. version v1.4.0. Swift do checkouts based on default branch tags.

PR #200, opened 2020-07-23
Increased deployment target to iOS 9 for Xcode 12 support

ttimpe:
Increased deployment target to iOS 9 for Xcode 12 support

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -6,7 +6,7 @@ let package = Package(
     name: "GTMSessionFetcher",
     platforms: [
         .macOS(.v10_10),
-        .iOS(.v8),
+        .iOS(.v9),
         .tvOS(.v9),
         .watchOS(.v2)
     ],


ttimpe:
@googlebot I signed it!

thomasvl:
What's the error/warning you hit?  I'm guessing this likely means we should update the podspec also to the same min version to support Xcode 12?

ttimpe:
Yes I guess the podspec needs to be updated as well. Xcode 12 only supports deployment targets higher or equal to iOS 9.

alak:
Any news on this ?

thomasvl:
Just waiting on an update PR that keeps the two packaging systems in sync.

karimhm:
I believe it is better to have 2 `Package.swift` manifest files, one for Swift tools version 5.3 and one for Swift tools version 5.2. Since the code is compatible with iOS versions earlier than 9.0.

thomasvl:
Actually, is this still needed after #202  But not listing anything, don't we just get the mins that Xcode wants to use?

thomasvl:
Closing this since I believe by removing the entry things are fixed now.

PR #201, opened 2020-08-27
Remove special character ç to prevent downstream failures in python3 encoding/decoding

liutenghou:
The character ç is causing downstream packages that use python3 to fail at the encoding/decoding step. Removing the character here seems like it would cause the lease side effects, and mitigate any issues that other users would have with a similar issue.

### DIFF
--- a/Source/GTMSessionFetcherLogging.m
+++ b/Source/GTMSessionFetcherLogging.m
@@ -226,7 +226,7 @@ static NSString *gLoggingProcessName = nil;
 }
 
 + (NSString *)htmlFileName {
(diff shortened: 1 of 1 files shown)


liutenghou:
@googlebot I signed it!

thomasvl:
python3 has unicode strings, and the leap from an ObjC library to some downstream python is a little strange.

My guess is whatever your python is reading things and not doing fully legit transforms on the unicode strings and that's what is causing you problems.

If we change this, we also risk breaking anyone that might have anything set up to collect logs from CI/Testing/etc.

PR #202, opened 2020-09-13
SwiftPM: Fix platform build warning

paulb777:
The platform specifications cause build warnings in Xcode 12. They're not necessary since gtm-session-fetcher supports all versions supported by Xcode versions that support Swift Package Manager.

Details in firebase/firebase-ios-sdk#6449, https://forums.swift.org/t/minimum-ios-version-xcode-12-and-build-warnings/40224, and https://github.com/apple/swift-evolution/blob/master/proposals/0236-package-manager-platform-deployment-settings.md

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -4,12 +4,6 @@ import PackageDescription
 
 let package = Package(
     name: "GTMSessionFetcher",
-    platforms: [
-        .macOS(.v10_10),
-        .iOS(.v8),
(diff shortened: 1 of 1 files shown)


thomasvl (approved):
I was gonna say that we need this to declare that we don't support Linux, but that never made it into the proposal.  So since we don't provide a different min, I guess this is fine.

Issue #218, opened 2020-11-12
+[GTMSessionUploadFetcher load] method takes too much time at app startup


maksymmalyhin:
Originally reported in https://github.com/firebase/firebase-ios-sdk/issues/6936.

<img width="2356" alt="Screen Shot 2020-11-10 at 11 09 33" src="https://user-images.githubusercontent.com/8662571/98623284-4fe4a700-2346-11eb-959b-031cf95767ca.png">

Minimal amount or ideally no code should be run in `+[GTMSessionUploadFetcher load]` as it contributes to pre-main app startup time.

mwyman:
Please check that you have not disabled GTMSESSION_RECONNECT_BACKGROUND_SESSIONS_ON_LAUNCH, which moves the background reloading out of `+load` and into an async block after the app launches.

maksymmalyhin:
@mwyman  Sorry for the delayed comment. 

It looks like the user defaults access from `+load` is guarded by the `GTMSESSION_RECONNECT_BACKGROUND_SESSIONS_ON_LAUNCH` in [GTMSessionUploadFetcher](https://github.com/google/gtm-session-fetcher/blob/91ed3d188eb95705fef3c249453b81f32dc557d1/Source/GTMSessionUploadFetcher.m#L152) (in contrast to [GTMSessionFetcher](https://github.com/google/gtm-session-fetcher/blob/91ed3d188eb95705fef3c249453b81f32dc557d1/Source/GTMSessionFetcher.m#L249)

PR #219, opened 2020-11-27
GTMSessionUploadFetcher: respect GTMSESSION_RECONNECT_BACKGROUND_SESSIONS_ON_LAUNCH flag

maksymmalyhin:
Fix https://github.com/google/gtm-session-fetcher/issues/218.

- replicate approach used in [GTMSessionFetcher](https://github.com/google/gtm-session-fetcher/blob/91ed3d188eb95705fef3c249453b81f32dc557d1/Source/GTMSessionFetcher.m#L249) to reduce app startup performance footprint.

@mwyman I didn't really do a lot of testing, so additional verification from your end is appreciated.

### DIFF
--- a/Source/GTMSessionUploadFetcher.m
+++ b/Source/GTMSessionUploadFetcher.m
@@ -19,6 +19,12 @@
 
 #import "GTMSessionUploadFetcher.h"
 
+#if TARGET_OS_OSX && GTMSESSION_RECONNECT_BACKGROUND_SESSIONS_ON_LAUNCH
+// To reconnect background sessions on Mac outside +load requires importing and linking
+// AppKit to access the NSApplicationDidFinishLaunching symbol.
+#import <AppKit/AppKit.h>
+#endif
+
 static NSString *const kGTMSessionIdentifierIsUploadChunkFetcherMetadataKey = @"_upChunk";
 static NSString *const kGTMSessionIdentifierUploadFileURLMetadataKey        = @"_upFileURL";
 static NSString *const kGTMSessionIdentifierUploadFileLengthMetadataKey     = @"_upFileLen";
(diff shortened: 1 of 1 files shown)


maksymmalyhin:
@mwyman I also have a question as for the default value of the flag `GTMSESSION_RECONNECT_BACKGROUND_SESSIONS_ON_LAUNCH ` defined [here](https://github.com/google/gtm-session-fetcher/blob/91ed3d188eb95705fef3c249453b81f32dc557d1/Source/GTMSessionFetcher.h#L437). What is the reason for the flag being set to `1` only for iOS 13+ devices? From the first glance it looks like it should be preferable behaviour for all iOS, tvOS and macOS devices.

mwyman:
> @mwyman I also have a question as for the default value of the flag `GTMSESSION_RECONNECT_BACKGROUND_SESSIONS_ON_LAUNCH ` defined [here](https://github.com/google/gtm-session-fetcher/blob/91ed3d188eb95705fef3c249453b81f32dc557d1/Source/GTMSessionFetcher.h#L437). What is the reason for the flag being set to `1` only for iOS 13+ devices? From the first glance it looks like it should be preferable behaviour for all iOS, tvOS, and macOS devices and probably watchOS as well (though may require additional code changes to support watchOS case).

Since this was a change in behavior, we gated it on using the latest target OS at the time to avoid an unexpected change in launch behavior; was considered (right or wrong) that projects that had targeted the latest OS at that time were staying more on-top of things.

GTMSessionFetcher.podspec
```diff
@@ -2,7 +2,7 @@
 # to import GTMSessionFetcher via the CocoaPods dependency Manager.
 Pod::Spec.new do |s|
   s.name        = 'GTMSessionFetcher'
-  s.version     = '1.5.0'
+  s.version     = '1.6.0'
```
mwyman, line 5: Please don't bump the version number in this PR. We like to keep that separate from individual code changes.

maksymmalyhin:
> Since this was a change in behavior, we gated it on using the latest target OS at the time to avoid an unexpected change in launch behavior; was considered (right or wrong) that projects that had targeted the latest OS at that time were staying more on-top of things.

I see, that makes sense. 

We would prefer to have it as a default for all Firebase users but it's a bit difficult (not sure if possible at all) to set up without an extra step for developers. I guess, the default behaviour may be updated in conjunction with the major the version bump which will indicate the braking change for the library users. I assume most of the users lock the dependency on a particular major version. WDYT?

mwyman (changes requested):

mwyman (approved):

PR #220, opened 2020-12-14
Fix Swift Package Manager support

woxtu:
Hi,

The current release installed via SPM cannot import headers as follows:

```objective-c
#import <GTMSessionFetcher/GTMSessionFetcher.h>
```

This PR creates umbrella directories to fix that.

### DIFF
--- a/Source/SwiftPackage/GTMSessionFetcher/GTMGatherInputStream.h
+++ b/Source/SwiftPackage/GTMSessionFetcher/GTMGatherInputStream.h
@@ -1,0 +1,1 @@
+../../GTMGatherInputStream.h
--- a/Source/SwiftPackage/GTMSessionFetcher/GTMMIMEDocument.h
+++ b/Source/SwiftPackage/GTMSessionFetcher/GTMMIMEDocument.h
@@ -1,0 +1,1 @@
+../../GTMMIMEDocument.h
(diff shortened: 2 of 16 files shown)


thomasvl:
@karimhm fyi

karimhm:
@thomasvl isn't symbolic links not desired because they are fragile?

thomasvl:
> @thomasvl isn't symbolic links not desired because they are fragile?

That was one thing I wasn't sure about, since it's limited to SwiftPM it avoid the general problem of the two locations being considered different files and the duplicate definitions one would otherwise get with them.

karimhm:
According to Apple's [Xcode 11.4 Release Notes](https://developer.apple.com/documentation/xcode-release-notes/xcode-11_4-release-notes?preferredLanguage=occ#Apple-Clang-Compiler):

> Xcode 11.4 turns the -Watimport-in-framework-header warning on by default. Using @import in framework headers is discouraged, because doing so requires all importers to use modules. (59222634) (FB7566046)

JonasGessner:
SPM is the go-to package manager in the Apple world now, not just for Swift but also ObjC. I don't just speak for myself when I say that Cocoapods and Carthage are on their way out. I appreciate that you have added SPM package manifests, but they have not worked once for me. Please move on from Cocoapods and please focus on getting all your libraries to actually work with SPM! This should be a top priority for these projects. GTMSessionFetcher, GTMAppAuth, the objc REST apis, etc.

wallyf1972 (approved):

thomasvl:
I'm hoping https://github.com/google/gtm-session-fetcher/pull/302 will resolve the issues you hit here.  It restructures things to simplify what has to go into the Package.swift.

PR #231, opened 2021-05-03
Some cleanup after running clang-format for PR #229.

mwyman:
- Re-formats `@synthesize` blocks wrapped in clang-format off/on comments, as we generally find them more readable one-per-line.
- Moves some blocks into locals first, to work around clang-format adding terrible indentation.

### DIFF
--- a/Source/GTMSessionFetcher.m
+++ b/Source/GTMSessionFetcher.m
@@ -1108,16 +1108,14 @@ NSData *_Nullable GTMDataFromInputStream(NSInputStream *inputStream, NSError **o
 
     // Simulate receipt of redirection.
     if (willRedirectBlock) {
-      [self invokeOnCallbackUnsynchronizedQueueAfterUserStopped:YES
-                                                          block:^{
-                                                            willRedirectBlock(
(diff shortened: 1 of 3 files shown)


Source/GTMSessionFetcher.m
```diff
@@ -1121,16 +1121,14 @@ - (void)simulateDataCallbacksForTestBlockWithBodyData:(NSData *GTM_NULLABLE_TYPE
...
-                                                          block:^{
-                                                            willRedirectBlock(
-                                                                (NSHTTPURLResponse *)response,
-                                                                self->_request,
-                                                                ^(NSURLRequest *redirectRequest){
-                                                                    // For simulation, we'll assume
-                                                                    // the app will just continue.
-                                                                });
-                                                          }];
+      __auto_type block = ^{
```
thomasvl, line 1124: Is __auto_type support in all our Xcode versions or should we bump that first?

mwyman: Good point, I'll take that on first.

thomasvl (approved):

PR #233, opened 2021-05-06
Remove GTMHTTPFetcher backwards compatibility code.

mwyman:
Compatibility with the old GTMHTTPFetcher is no longer necessary. Removing all references and compatibility code.

### DIFF
--- a/Source/GTMSessionFetcher.h
+++ b/Source/GTMSessionFetcher.h
@@ -179,9 +179,8 @@
 // Note: cookies set while following redirects will be sent to the server, as
 // the redirects are followed by the fetcher.
 //
-// To completely disable cookies, similar to setting cookieStorageMethod to
-// kGTMHTTPFetcherCookieStorageMethodNone, adjust the session configuration
-// appropriately in the fetcher or fetcher service:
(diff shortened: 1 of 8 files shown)


Source/GTMSessionFetcher.h
```diff
@@ -624,7 +580,6 @@ NSData *_Nullable GTMDataFromInputStream(NSInputStream *inputStream, NSError **o
 - (nullable id<NSURLSessionDelegate>)sessionDelegate;
 - (nullable NSDate *)stoppedAllFetchersDate;
 
-// Methods for compatibility with the old GTMHTTPFetcher.
 @property(atomic, readonly, strong, nullable) NSOperationQueue *delegateQueue;
```
thomasvl, line 628: This will take some checking for a future cl, but we likely should consider pruning these.

thomasvl (approved):

Source/GTMSessionFetcher.h
mwyman: Yeah, I wasn't confident enough in removing that one for this PR. Will take a deeper look later; figured it was harmless enough for now.

PR #237, opened 2021-06-10
Add the `swift_async(none)` attribute to calm `-Wcompletion-handler` warnings.

mwyman:
Xcode 13/Clang 13 introduces the `-Wcompletion-handler` warning on methods that do not appear to call the completion handler. The SDKs add an `NS_SWIFT_DISABLE_ASYNC` macro to add the attribute to quash this, but rather than depend on the SDK at this time, check the availability of the underlying `swift_async` attribute from clang and use our own macro.

The methods being annotated are: an `URLSession` delegate method that sends the completion to a callback block (else calls the completion), but the compiler doesn't catch this correctly; an `UIApplication` delegate method for handling background URLSession events, and does not call the completion handler if creating a fetcher to handle them fails - not sure the correct handling for this case, but seems like dropping the events may not be the right approach.


Source/GTMSessionFetcher.m
```diff
@@ -134,6 +134,11 @@
 #define GTM_SDK_REQUIRES_SECTRUSTEVALUATEWITHERROR 0
 #endif
 
+#if __has_attribute(swift_async)
+// Once Clang 13/Xcode 13 can be assumed, can switch to NS_SWIFT_DISABLE_ASYNC.
+#define GTM_SWIFT_DISABLE_ASYNC __attribute__((swift_async(none)))
+#endif
```
thomasvl, line 140: You need the else to define it away, no?

mwyman: Oops, yep. That's what happens when I switch from defining the NS_SWIFT_… symbol as empty if it's not defined to defining my own thing… gotta remember the else!

thomasvl (approved):

PR #242, opened 2021-06-23
Clean up backwards-compatibility code and add deprecations.

mwyman:
This reverts commit 0d539fdd85c004076df09790d631cc29e22d4dd3.

Re-removes the backwards compatibility code. Next release should be v2.0.0.

### DIFF
--- a/Source/GTMSessionFetcher.h
+++ b/Source/GTMSessionFetcher.h
@@ -179,9 +179,8 @@
 // Note: cookies set while following redirects will be sent to the server, as
 // the redirects are followed by the fetcher.
 //
-// To completely disable cookies, similar to setting cookieStorageMethod to
-// kGTMHTTPFetcherCookieStorageMethodNone, adjust the session configuration
-// appropriately in the fetcher or fetcher service:
+// To completely disable cookies, adjust the session configuration appropriately
+// in the fetcher or fetcher service:
 //  fetcher.configurationBlock = ^(GTMSessionFetcher *configFetcher,
 //                                 NSURLSessionConfiguration *config) {
 //    config.HTTPCookieAcceptPolicy = NSHTTPCookieAcceptPolicyNever;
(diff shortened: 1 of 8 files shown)


Source/GTMSessionFetcher.h
```diff
@@ -384,45 +383,6 @@
...
 #endif
 
-#if (TARGET_OS_TV || TARGET_OS_WATCH ||                          \
-     (!TARGET_OS_IPHONE && defined(MAC_OS_X_VERSION_10_11) &&    \
-      MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_X_VERSION_10_11) || \
-     (TARGET_OS_IPHONE && defined(__IPHONE_9_0) &&               \
-      __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_9_0))
-#ifndef GTM_USE_SESSION_FETCHER
-#define GTM_USE_SESSION_FETCHER 1
-#endif
```
thomasvl, line 394: Given the min versions, I think this can reduce to just a #ifndef...#define...#endif

Source/GTMSessionFetcher.h
```diff
@@ -384,45 +383,6 @@
...
-#endif
-
-#if !defined(GTMBridgeFetcher)
```
thomasvl, line 397: And keep this block also.

Source/GTMSessionFetcher.h
```diff
@@ -1234,12 +1185,6 @@ NSData *_Nullable GTMDataFromInputStream(NSInputStream *inputStream, NSError **o
 
 @end
 
-@interface GTMSessionFetcher (BackwardsCompatibilityOnly)
-// Clients using GTMSessionFetcher should set the cookie storage explicitly themselves.
-// This method is just for compatibility with the old GTMHTTPFetcher class.
-- (void)setCookieStorageMethod:(NSInteger)method;
```
thomasvl, line 1240: keep this, but also tag it as deprecated?

Source/GTMSessionFetcher.h
```diff
@@ -633,10 +593,6 @@ NSData *_Nullable GTMDataFromInputStream(NSInputStream *inputStream, NSError **o
 }  // extern "C"
 #endif
 
-#if !GTM_USE_SESSION_FETCHER
```
thomasvl, line 636: I _think_ this and the conditional definition on the `fetcherService` property is likely safe to since the on things that seem like they could break would be someone explicitly using the SessionFetcher but with the `#define` for the bridge macros set to `0`, and that seems like it an odd state.

thomasvl (commented):
How about doing a modified subset of this?  We could land this and call it a 2.0, but until GTMAppAuth gets a release, no one would be able to update it, which could cause problems if we have other updates/bug fixes in the future.

Source/GTMSessionFetcher.h
mwyman: Done.

Source/GTMSessionFetcher.h
mwyman: Done.

Source/GTMSessionFetcher.h
mwyman: Removed the GTM_USE_SESSION_FETCHER altogether, since I don't believe there's a point to support it being defined to 0 at this point.

Source/GTMSessionFetcher.h
mwyman: Agreed. I don't think setting GTM_USE_SESSION_FETCHER=0 is really a viable configuration at this point.

thomasvl (approved):

PR #247, opened 2021-07-27
Only wrap the callback queue if callbacks are callbacks are provided that expect serialization.

mwyman:
Accumulate data and progress callbacks could reasonably expect to require serialization; otherwise it's unnecessary to wrap the dispatch queue in case the client provided a concurrent queue, and clients should not have such unnecessary queues created (and most provide serial queues making the wrapping unnecessary anyways).

### DIFF
--- a/Source/GTMSessionFetcher.m
+++ b/Source/GTMSessionFetcher.m
@@ -229,8 +229,8 @@ static GTMSessionFetcherTestBlock _Nullable gGlobalTestBlock;
                            // released)
   id _userData;            // retained, if set by caller
   NSMutableDictionary *_properties;  // more data retained for caller
-  dispatch_queue_t _callbackQueue;   // the target of _serialCallbackQueue
-  dispatch_queue_t _serialCallbackQueue;
+  dispatch_queue_t _callbackQueue;   // the target of _internalCallbackQueue
+  dispatch_queue_t _internalCallbackQueue;
   dispatch_group_t _callbackGroup;   // read-only after creation
   NSOperationQueue *_delegateQueue;  // immutable after beginFetch
 
@@ -528,6 +528,9 @@ static GTMSessionFetcherTestBlock _Nullable gGlobalTestBlock;
     // This ivar is set only here on the initial beginFetch so need not be synchronized.
     _initialBeginFetchDate = [[NSDate alloc] init];
   }
+
+  // Lock in the internal callback queue the first time beginFetch is called.
+  (void)[self internalCallbackQueue];
 
   if (self.sessionTask != nil) {
     // If cached fetcher returned through fetcherWithSessionIdentifier:, then it's
@@ -1786,24 +1789,24 @@ NSData *_Nullable GTMDataFromInputStream(NSInputStream *inputStream, NSError **o
   // the blocks being released may call back into the fetcher or fetcher
   // service.
   dispatch_queue_t NS_VALID_UNTIL_END_OF_SCOPE holdCallbackQueue;
-  dispatch_queue_t NS_VALID_UNTIL_END_OF_SCOPE holdSerialCallbackQueue;
+  dispatch_queue_t NS_VALID_UNTIL_END_OF_SCOPE holdInternalCallbackQueue;
   GTMSessionFetcherCompletionHandler NS_VALID_UNTIL_END_OF_SCOPE holdCompletionHandler;
   @synchronized(self) {
     GTMSessionMonitorSynchronized(self);
 
     holdCallbackQueue = _callbackQueue;
-    holdSerialCallbackQueue = _serialCallbackQueue;
+    holdInternalCallbackQueue = _internalCallbackQueue;
     holdCompletionHandler = _completionHandler;
 
     _callbackQueue = nil;
-    _serialCallbackQueue = nil;
+    _internalCallbackQueue = nil;
     _completionHandler = nil;  // Setter overridden in upload. Setter assumed to be used externally.
   }
 
   // Set local callback pointers to nil here rather than let them release at the end of the scope
   // to make any problems due to the blocks being released be a bit more obvious in a stack trace.
   holdCallbackQueue = nil;
-  holdSerialCallbackQueue = nil;
+  holdInternalCallbackQueue = nil;
   holdCompletionHandler = nil;
 
   self.configurationBlock = nil;
@@ -1903,11 +1906,11 @@ NSData *_Nullable GTMDataFromInputStream(NSInputStream *inputStream, NSError **o
           void (^resumeBlock)(NSData *) = _resumeDataBlock;
           _resumeDataBlock = nil;
 
-          // Save serialCallbackQueue since releaseCallbacks clears it.
-          dispatch_queue_t serialCallbackQueue = [self serialCallbackQueueUnsynchronized];
+          // Save internalCallbackQueue since releaseCallbacks clears it.
+          dispatch_queue_t internalCallbackQueue = [self internalCallbackQueueUnsynchronized];
           dispatch_group_enter(_callbackGroup);
           [(NSURLSessionDownloadTask *)oldTask cancelByProducingResumeData:^(NSData *resumeData) {
-            [self invokeOnCallbackQueue:serialCallbackQueue
+            [self invokeOnCallbackQueue:internalCallbackQueue
                        afterUserStopped:YES
                                   block:^{
                                     resumeBlock(resumeData);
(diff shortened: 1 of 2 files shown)


Source/GTMSessionFetcher.m
```diff
@@ -3749,21 +3753,30 @@ - (void)setCallbackQueue:(nullable dispatch_queue_t)queue {
...
+- (nonnull dispatch_queue_t)internalCallbackQueueUnsynchronized {
   GTMSessionCheckSynchronized(self);
 
-  if (!_serialCallbackQueue) {
-    _serialCallbackQueue = SerialCallbackQueueForTargetQueue(_callbackQueue);
+  if (!_internalCallbackQueue) {
+    // If the fetcher was provided blocks for passive sequential callbacks (not ones where the
+    // system waits for a response, such as the response callback), create a serial queue wrapper
+    // targeting the client-provided callback queue.
+    bool hasSerialCallbacks = self.accumulateDataBlock || self.sendProgressBlock ||
```
thomasvl, line 3771: The only other edge case I can think of is retries.  Is there any way retries could get into trouble with callbacks showing up out of order?

Source/GTMSessionFetcher.m
```diff
@@ -3749,21 +3753,30 @@ - (void)setCallbackQueue:(nullable dispatch_queue_t)queue {
...
-- (nonnull dispatch_queue_t)serialCallbackQueueUnsynchronized {
+- (nonnull dispatch_queue_t)internalCallbackQueueUnsynchronized {
   GTMSessionCheckSynchronized(self);
 
-  if (!_serialCallbackQueue) {
-    _serialCallbackQueue = SerialCallbackQueueForTargetQueue(_callbackQueue);
+  if (!_internalCallbackQueue) {
+    // If the fetcher was provided blocks for passive sequential callbacks (not ones where the
+    // system waits for a response, such as the response callback), create a serial queue wrapper
+    // targeting the client-provided callback queue.
```
thomasvl, line 3770: Maybe expand this comment that we do this to help out developers by ensuring they won't have to deal with out of order messages if they failed to make their custom queue serial.

Source/GTMSessionFetcher.m
```diff
@@ -3749,21 +3753,30 @@ - (void)setCallbackQueue:(nullable dispatch_queue_t)queue {
...
   }  // @synchronized(self)
 }
 
-- (nonnull dispatch_queue_t)serialCallbackQueueUnsynchronized {
+- (nonnull dispatch_queue_t)internalCallbackQueueUnsynchronized {
   GTMSessionCheckSynchronized(self);
 
-  if (!_serialCallbackQueue) {
-    _serialCallbackQueue = SerialCallbackQueueForTargetQueue(_callbackQueue);
+  if (!_internalCallbackQueue) {
```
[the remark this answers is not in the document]
mwyman: Since `_internalCallbackQueue` is what we're using for the actual dispatches, the `_callbackQueue`, whatever it is, needs to get assigned to it when it's the main queue. I can move the check into here and just have the serial call do the wrapping explicitly.

Source/GTMSessionFetcher.m
mwyman: I'd been contemplating that, but the retry block is called after the NSURLSessionTask completes with an error; it'll only be called once at at time, and any further action on the fetcher waits until the response block is called, which should provide its own effective serialization.

I guess there is a question about the `resumeData` block, which could be dispatched immediately before the completion block. I'll add that to this check.

Source/GTMSessionFetcher.m
mwyman: Done.

thomasvl (approved):

fumoboy007:
Hmm… shouldn’t we dispatch serially in more cases? For example, if *any* block is set, we should probably ensure that the completion handler is called after those blocks are finished.

If performance impact is the concern, have we tested the impact?

thomasvl:
> Hmm… shouldn’t we dispatch serially in more cases? For example, if _any_ block is set, we should probably ensure that the completion handler is called after those blocks are finished.

Lost me, how is that not what this CL does?

Unless you are referring to some code path where even if the callback queue were serial, we could get fire progress/etc. callbacks before the completion one?  If that's the case, it's a different bug and the insertion of the serial queue does nothing to prevent it.

> If performance impact is the concern, have we tested the impact?

Most of these isn't about performance at all, it is about concerns for the overhead of a queue being inserted for *every* fetcher that is alive.  Yes, Apple says they are cheap, but that doesn't mean they are free.  Fetchers can/are used in iOS extensions and watchOS apps which have much more limited resources than other contexts, so one has to keep that in mind.

And all of this serial queue insertion is really trying to protect developers from what could also be considered their own mistake - them not using a serial queue when their callbacks assume serial calling order.  A concurrent queue for multiple fetchers where they only use completion calls back is fine/correct. With the changes we've now moved in a direction where we're putting a more memory expense for those developers, and the expense isn't doing anything for them since they already were correct.  Likewise, for the developers that did want some of these other callbacks, if they correctly used a serial queue already, we're just inserting another one that buys them nothing but some extra expense.

fumoboy007:
> > Hmm… shouldn’t we dispatch serially in more cases? For example, if _any_ block is set, we should probably ensure that the completion handler is called after those blocks are finished.
> 
> Lost me, how is that not what this CL does?

I see `didReceiveResponseBlock` and other blocks are not included in the heuristic. Shouldn’t we also guarantee that these will finish before the completion block is dispatched?

> Most of these isn't about performance at all, it is about concerns for the overhead of a queue being inserted for _every_ fetcher that is alive. Yes, Apple says they are cheap, but that doesn't mean they are free. Fetchers can/are used in iOS extensions and watchOS apps which have much more limited resources than other contexts, so one has to keep that in mind.

Right, that’s what I meant by performance (CPU + memory). Perhaps efficiency is more accurate.

> And all of this serial queue insertion is really trying to protect developers from what could also be considered their own mistake - them not using a serial queue when their callbacks assume serial calling order. A concurrent queue for multiple fetchers where they only use completion calls back is fine/correct. With the changes we've now moved in a direction where we're putting a more memory expense for those developers, and the expense isn't doing anything for them since they already were correct. Likewise, for the developers that did want some of these other callbacks, if they correctly used a serial queue already, we're just inserting another one that buys them nothing but some extra expense.

Isn’t this the same argument as whether to add an assertion to catch programmer error vs. leave the assertion out for more efficiency? That’s why I suggested measuring the efficiency impact so that we can see what the actual trade-off is.

My instinct was that (a) it’s really easy to make this mistake, (b) it’s really hard to find this mistake, and (c) the efficiency impact of an intermediate queue is probably negligible, so I thought that adding this protection would be a good trade-off.

thomasvl:
The same measurement argument likely should have been made on the initial PR.  😀 

There's still some value in trying to figure out if this cost can be removed in as many cases as possible; which is part of what these PRs are trying to do.  Yes, there maybe other blocks that should be checked.

> Isn’t this the same argument as whether to add an assertion to catch programmer error vs. leave the assertion out for more efficiency?

Assertions can/should be compiled away in release code.  The library does generally do some extra checks in Debug builds to help catch mistakes for just these sorta reasons.  But this isn't something that only comes into play in a Debug build.

From past experience on other code, we're a little concerned about developers (that are doing their queues/callbacks correctly) having Xcode suddenly showing them a potential non trivial number of extra queues when their app is doing networking.  So there is some value in trying to strike the right balance here in helping someone that is making some mistakes without taxing folks that are doing things correctly.  At the moment I'm not aware of any of Apple's api that do similar things (or third party ones); so it seems worth ensuring this is a good tradeoff before it is included in a release.

mwyman:
> I see `didReceiveResponseBlock` and other blocks are not included in the heuristic. Shouldn’t we also guarantee that these will finish before the completion block is dispatched?

Some blocks, like `didReceiveResponseBlock`, must call the block passed to them as a parameter before the fetch continues, so there is baked in serialization already.

fumoboy007:
> The same measurement argument likely should have been made on the initial PR. 😀

I created a couple [benchmarks](https://github.com/google/gtm-session-fetcher/files/7018511/TargetQueueBenchmark.zip) to measure the latency overhead and the memory overhead of an intermediate queue. The results are below. (They are device-specific results but we can use them to grasp the general magnitude of the overhead.)

| Type | Overhead |
| --- | --- |
| Latency | 5 ns |
| Memory | 72 B |

I think this level of overhead is negligible, so I think adding this protection (for all callbacks) is a great trade-off.

PR #248, opened 2021-07-28
New APIs -[GTMSessionFetcherService addHeaderDecorator:] and -removeHeaderDecorator:

### LINKED ISSUE #246: [GTMSessionFetcherService] Allow decorating HTTP headers per request
Currently, the only way to globally customize HTTP headers per request is to subclass `GTMSessionFetcherService` and override `-fetcherWithRequest:fetcherClass:`. (This makes things awkward if you want to inject a unique HTTP header value into every request, but someone else has already subclassed `GTMSessionFetcherService`).

It is possible today to provide a custom `NSURLSessionConfiguration *` (which has a property `HTTPAdditionalHeaders` to provide a dictionary of headers to be added to a request) via `-[GTMSessionFetcherService setConfigurationBlock]`'. However, these headers are only applied at the *start* of a session; multiple `GTMSessionFetcher` instances can and will be dispatched through the life of a single session, so they all share the same `HTTPAdditionalHeaders`. That means it's not possible to use this to separately customize the HTTP headers for each fetcher with this solution.

This feature request tracks adding a new API to `GTMSessionFetcherService` to register/unregister "decorators" which add HTTP header(s) to a request before it starts.

Strawman:

```
@protocol GTMSessionFetcherHeaderDecorator <NSObject>
- (nullable NSDictionary<id, id> *)additionalHeadersForRequest:(nonnull NSURLRequest *)request;
@end

@interface GTMSessionFetcherService : NSObject<GTMSessionFetcherServiceProtocol>
// Holds a strong reference to `decorator`. If multiple decorators add the same header to a request,
// the most recent decorator passed to this method wins.
- (void)addHeaderDecorator:(id<GTMSessionFetcherHeaderDecorator>)decorator;
- (void)removeHeaderDecorator:(id<GTMSessionFetcherHeaderDecorator>)decorator;
@end
```

It's also possible for us to hold `weak` references to the `decorator`s, but I'm not aware of a solution in either Foundation or CoreFoundation for an ordered collection of `weak` references, so we'd have to give up the "most recently added decorator wins" idea if we do that.

bhamiltoncx:
Fixes #246.

Tests included.


Source/UnitTests/GTMSessionFetcherServiceTest.m
```diff
@@ -44,6 +44,46 @@
   [self waitForExpectations:@[ fetcherStartedExpectation__, fetcherStoppedExpectation__ ] \
                     timeout:5.0];
 
+NSDictionary<NSString *, NSString *> *FetcherHeadersWithoutUserAgent(GTMSessionFetcher *fetcher) {
```
mwyman, line 47: `static`

Source/GTMSessionFetcherService.m
```diff
@@ -202,12 +208,13 @@ - (id)fetcherWithRequest:(NSURLRequest *)request fetcherClass:(Class)fetcherClas
   fetcher.skipBackgroundTask = self.skipBackgroundTask;
 #endif
 
+  [self decorateHeadersForRequest:request fetcher:fetcher];
```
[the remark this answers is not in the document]
mwyman: Could the client potentially want to provide new headers for a redirect, or on a retry?

Source/GTMSessionFetcherService.m
```diff
@@ -224,6 +231,54 @@ - (GTMSessionFetcher *)fetcherWithURLString:(NSString *)requestURLString {
...
+        needsCompact = YES;
+        continue;
+      }
+      NSDictionary<NSString *, NSString *> *_Nullable headers = [headerDecorator additionalHeadersForRequest:request];
+      for (NSString *headerName in headers) {
+        NSString *headerValue = headers[headerName];
+        [fetcher setRequestValue:headerValue forHTTPHeaderField:headerName];
+      }
+    }
+    if (needsCompact) {
```
mwyman, line 273: I'm curious how much of a concern this is likely to be. I would imagine in most cases the decorator is intended to live the entire duration of the service, and the reason for using weak pointers is (I think) largely to avoid a retain cycle.

I would foresee most fetcher service instances will have 0, 1, or 2 decorators, and I imagine most decorators will have lifetimes largely the same as the service itself.

Source/GTMSessionFetcherService.m
```diff
@@ -224,6 +231,54 @@ - (GTMSessionFetcher *)fetcherWithURLString:(NSString *)requestURLString {
...
+    }
+  }
+}
+
+- (void)decorateHeadersForRequest:(NSURLRequest *)request
+                          fetcher:(GTMSessionFetcher *)fetcher {
+  @synchronized(self) {
+    BOOL needsCompact = NO;
+    for (id<GTMSessionFetcherHeaderDecorator> headerDecorator in _headerDecorators) {
+      if (!headerDecorator) {
```
mwyman, line 261: I'm a bit leery calling into unknown external code while inside an `@synchronized` block. I wonder if it wouldn't be better to fetch `_headerDecorators.allObjects` into a local array and do the processing outside the synchronized block?

Source/GTMSessionFetcherService.m
```diff
@@ -156,6 +160,8 @@ - (instancetype)init {
     (TARGET_OS_IPHONE && defined(__IPHONE_9_0) && __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_9_0)
     _userAgent = GTMFetcherStandardUserAgentString(nil);
 #endif
+
+    _headerDecorators = [NSPointerArray weakObjectsPointerArray];
```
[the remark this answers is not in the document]
bhamiltoncx: Sure, done.

Source/GTMSessionFetcherService.m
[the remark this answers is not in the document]
bhamiltoncx: Sure. Added an enum `GTMSessionFetcherHeaderDecoratorPhase` with values for `Creation`, `Redirect`, and `Retry` that allows decorators to add headers at any of those steps of the fetcher.

Source/GTMSessionFetcherService.m
```diff
@@ -224,6 +231,54 @@ - (GTMSessionFetcher *)fetcherWithURLString:(NSString *)requestURLString {
...
+      [_headerDecorators removePointerAtIndex:i];
+    }
+  }
+}
+
+- (void)decorateHeadersForRequest:(NSURLRequest *)request
+                          fetcher:(GTMSessionFetcher *)fetcher {
+  @synchronized(self) {
+    BOOL needsCompact = NO;
+    for (id<GTMSessionFetcherHeaderDecorator> headerDecorator in _headerDecorators) {
```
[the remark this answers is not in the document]
bhamiltoncx: No need to change here, as fast iteration on `nil` is a no-op.

Source/UnitTests/GTMSessionFetcherServiceTest.m
bhamiltoncx: Done, thanks.

Source/GTMSessionFetcherService.m
bhamiltoncx: I removed the compact logic to simplify it. If folks are adding hundreds of decorators and letting the weak references go nil, we can revisit this.

Source/GTMSessionFetcherService.m
bhamiltoncx: Sure, done.

bhamiltoncx:
Hmm, there's one failing test (swiftpm debug), but I don't see how my changes could have affected it:

https://github.com/google/gtm-session-fetcher/pull/248/checks?check_run_id=3195657631

thomasvl:
> Hmm, there's one failing test (swiftpm debug), but I don't see how my changes could have affected it:

FYI - went away on a rerun, so likely was flake. ☹️

bhamiltoncx:
> FYI - went away on a rerun, so likely was flake. ☹️

Thanks for confirming it wasn't this PR!

Source/GTMSessionFetcher.h
```diff
@@ -731,6 +744,14 @@ NSData *_Nullable GTMDataFromInputStream(NSInputStream *inputStream, NSError **o
 // affect a fetch after the fetch has begun.
 - (void)setRequestValue:(nullable NSString *)value forHTTPHeaderField:(NSString *)field;
 
+// Holds a weak reference to `decorator`. Each registered `decorator` can add HTTP header(s) to the
+// request before it starts.  If multiple decorators add the same header to a request, the most
+// recent decorator passed to this method wins.
+- (void)addHeaderDecorator:(id<GTMFetcherHeaderDecoratorProtocol>)decorator;
```
mwyman, line 750: Is this useful to add on a per-fetcher basis? When creating an individual, standalone fetcher it would be sufficient to directly add the desired headers. I would think the main utility of the decorator would be when used with the service, where we want all fetchers created with the service to use the decorators.

From the API surface of a GTMSessionFetcher instance, I wonder if the only thing we might want to expose would be a way to explicitly skip using the service's header decorator (if, for some reason, this was desired on an individual fetcher?).

This would allow us not to need to pass decorators from the service to the individual fetcher.

Source/GTMSessionFetcher.m
```diff
@@ -1689,8 +1704,72 @@ - (void)authorizer:(id<GTMFetcherAuthorizationProtocol>)auth
...
+  }
+}
+
+- (void)decorateHeadersForRequest:(NSURLRequest *)fetchRequest
+                 headerDecorators:(NSArray<id<GTMFetcherHeaderDecoratorProtocol>> *)headerDecorators {
+  GTMSessionCheckNotSynchronized(self);
+
+  GTMSESSION_LOG_DEBUG(@"GTMSessionFetcher decorateHeaders start %zu decorators", headerDecorators.count);
+
+  dispatch_queue_t decorateCompletionQueue;
```
mwyman, line 1717: The authorization protocol provides the means to stop authorization, if necessary (`stopAuthorizationForRequest:`). Given the implementation here, it seems it would be difficult to implement a similar scheme in case a decorator takes too much time and a fetch is stopped, no?

Source/GTMSessionFetcher.m
```diff
@@ -819,6 +821,19 @@ - (void)beginFetchMayDelay:(BOOL)mayDelay
     }
   }
 
+  if (mayDecorate && _headerDecorators.count) {
```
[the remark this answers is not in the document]
bhamiltoncx: Yeah, this method is full of existing race conditions — it seems like everything assumes the fetcher is (externally) immutable once this method starts.

Some existing race conditions:

1) https://github.com/google/gtm-session-fetcher/blob/fb94199ed54cbbf3b1f46de16df00b7da2bf0e5b/Source/GTMSessionFetcher.m#L642
2) https://github.com/google/gtm-session-fetcher/blob/fb94199ed54cbbf3b1f46de16df00b7da2bf0e5b/Source/GTMSessionFetcher.m#L659
3) https://github.com/google/gtm-session-fetcher/blob/fb94199ed54cbbf3b1f46de16df00b7da2bf0e5b/Source/GTMSessionFetcher.m#L784

There's [a comment that says the *request* is treated as immutable once the method starts](https://github.com/google/gtm-session-fetcher/blob/fb94199ed54cbbf3b1f46de16df00b7da2bf0e5b/Source/GTMSessionFetcher.m#L502) — maybe the intent is that the entire *fetcher* is treated as immutable once -beginFetch:... starts?

Source/GTMSessionFetcher.m
```diff
@@ -1689,8 +1704,72 @@ - (void)authorizer:(id<GTMFetcherAuthorizationProtocol>)auth
     @synchronized(self) {
       _request = authorizedRequest;
     }
-    [self beginFetchMayDelay:NO mayAuthorize:NO];
+    [self beginFetchMayDelay:NO mayAuthorize:NO mayDecorate:YES];
+  }
+}
+
+- (void)decorateHeadersForRequest:(NSURLRequest *)fetchRequest
+                 headerDecorators:(NSArray<id<GTMFetcherHeaderDecoratorProtocol>> *)headerDecorators {
```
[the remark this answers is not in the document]
bhamiltoncx: Thanks, done.

Source/GTMSessionFetcher.m
bhamiltoncx: Yeah, that's the downside of using blocks. However, the worst thing that happens is the header calculation result is thrown away when the decorator invokes its callback, right?

IMHO, the benefits of blocks (type safety, etc.) outweigh the benefits of allowing early cancellation of calculating headers. What do you think?

Source/GTMSessionFetcher.h
bhamiltoncx: That's true, the fetcher can just reach into the service directly. Thanks, will fix.

Source/GTMSessionFetcher.m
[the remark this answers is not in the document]
bhamiltoncx: Resolved by moving header decorators to the service.

Source/GTMSessionFetcher.h
```diff
@@ -608,6 +611,22 @@ NSData *_Nullable GTMDataFromInputStream(NSInputStream *inputStream, NSError **o
 }  // extern "C"
 #endif
 
+// Adds HTTP header(s) to a request before it's sent out. See `-[GTMSessionFetcherService
+// addHeaderDecorator:]` and `-[GTMSessionFetcherService removeHeaderDecorator:]`.
+@protocol GTMFetcherHeaderDecoratorProtocol <NSObject>
+
+// Return YES if this request should have its headers decorated, NO otherwise.
+- (BOOL)shouldDecorateHeadersForRequest:(NSURLRequest *)request;
```
mwyman, line 619: We should probably comment queue/threading expectations for these methods. For example, neither should make I/O calls that could block, since both are potentially called from the main queue.

Source/GTMSessionFetcher.m
```diff
@@ -1689,10 +1696,79 @@ - (void)authorizer:(id<GTMFetcherAuthorizationProtocol>)auth
...
+  BOOL needsDecorate = NO;
+  for (id<GTMFetcherHeaderDecoratorProtocol> decorator in headerDecorators) {
+    if (![decorator shouldDecorateHeadersForRequest:fetchRequest]) {
+      ++headerDecoratorsIndex;
+      continue;
+    }
+    if (!needsDecorate) {
+      needsDecorate = YES;
+      decorateCompletionQueue =
+          dispatch_queue_create("com.google.GTMFetcherHeaderDecorator", DISPATCH_QUEUE_SERIAL);
```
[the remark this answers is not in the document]
bhamiltoncx: Discussed offline — no need to target since this is a private implementation detail. If profiling reveals a bottleneck, we can revisit.

Source/GTMSessionFetcher.h
bhamiltoncx: Thanks, added documentation on threading contract.

bhamiltoncx:
Generalized to support decorators that can modify the NSURLRequest at request start, retry, and finish time.

Source/GTMSessionFetcher.h
```diff
@@ -608,6 +610,51 @@ NSData *_Nullable GTMDataFromInputStream(NSInputStream *inputStream, NSError **o
...
+// After the decorator's work is complete, the decorator must invoke `handler(request, error)`
+// either synchronously or asynchronously (on any queue).
+//
+// If `error` is non-nil, then the fetcher is stopped with the given error, and any further
+// decorators' `-fetcherWillStart:completionHandler:` methods are not invoked.
+//
+// Otherwise, the decorator may use `[fetcher.request mutableCopy]`, make changes to the mutable
+// copy of the request, and pass the result to the handler via the `request` parameter.
+//
+// To distinguish the initial fetch from retries, the decorator can look at `fetcher.retryCount`.
```
mwyman, line 633: Awesome! I was just thinking about this being a way to distinguish, and am glad you're calling it out! 👍🏻

Source/GTMSessionFetcher.h
```diff
@@ -608,6 +610,51 @@ NSData *_Nullable GTMDataFromInputStream(NSInputStream *inputStream, NSError **o
...
+// Invoked just before a fetcher's request starts.
+//
+// After the decorator's work is complete, the decorator must invoke `handler(request, error)`
+// either synchronously or asynchronously (on any queue).
+//
+// If `error` is non-nil, then the fetcher is stopped with the given error, and any further
+// decorators' `-fetcherWillStart:completionHandler:` methods are not invoked.
+//
+// Otherwise, the decorator may use `[fetcher.request mutableCopy]`, make changes to the mutable
+// copy of the request, and pass the result to the handler via the `request` parameter.
```
mwyman, line 631: We should document what happens if the `request` parameter to the callback is `nil`. Perhaps something along the lines of "If no changes to the request are being made, pass `nil` to the callback for the `request` parameter." (happy to have that word-smithed)?

Source/GTMSessionFetcher.m
```diff
@@ -1689,10 +1698,93 @@ - (void)authorizer:(id<GTMFetcherAuthorizationProtocol>)auth
...
     }
-    [self beginFetchMayDelay:NO mayAuthorize:NO];
+    [self beginFetchMayDelay:NO mayAuthorize:NO mayDecorate:YES];
   }
 }
 
+- (void)applyDecoratorsAtRequestWillStart:(NSArray<id<GTMFetcherDecoratorProtocol>> *)decorators
+                          startingAtIndex:(NSUInteger)index {
+  GTMSessionCheckNotSynchronized(self);
+  if (index == decorators.count) {
```
mwyman, line 1708: `>=`, just for caution's sake?

(Same comment in the other `apply…` method)

Source/GTMSessionFetcher.h
bhamiltoncx: Thanks, done.

Source/GTMSessionFetcher.m
bhamiltoncx: Sure! Done.

mwyman (approved):

PR #250, opened 2021-08-05
Update update_version.py

mrprogrammer2938:


mwyman:
This change is unnecessary; the script is meant to be run manually and omitting the argument will be readily apparent.

PR #256, opened 2021-09-14
Improve GTMSessionFetcher logging

mega210:
It appears that GTMSession that failed due to request timeout does not have basic useful information at logs.

For example in real world if session is failed by timeout there is no start time and URL associated with the session in the logs. There is no such problem in case where response started to receiving (i.e. header was received). Because events didReceiveResponse and willCacheResponse both put URL in the log.


mega210:
> Having said that, where this is being added, it looks like it will generate _a lot_ more logging.

I don't think it will be _a lot_ more, because didReceiveResponse and willCacheResponse are already logging similar information. My concern is in case of "timeout" scenario request completely disappearing from logs.

mega210:
Yes, you're right. `GTM_LOG_SESSION_DELEGATE` name kinda misleading though. Maybe this is a good idea to rename it with something like `GTMSESSION_LOG_DEBUG_VERBOSE` along with this changes?

mega210:
Done

PR #257, opened 2021-09-30
Add explicit concurrent callback queue support for the fetcher service.

mwyman:
Clients that wish to provide a concurrent queue for the service's callback queue can call the new `setConcurrentCallbackQueue:` method, and each new fetcher instance will be provided a new serial queue targeting the concurrent callback queue; this will ensure callbacks for each instance are executed in order, while callbacks from separate fetcher instances are not blocked by each other.

If the callback queue is set directly via the `.callbackQueue` property, the service provides that callback queue directly to all new fetcher instances (this is current behavior).

Setting a fetcher service's callback queue after the service has been used to create fetcher instances is not guaranteed to behave in a specific manner, and should be considered an error.

### DIFF
--- a/Source/GTMSessionFetcherService.h
+++ b/Source/GTMSessionFetcherService.h
@@ -109,9 +109,10 @@ extern NSString *const kGTMSessionFetcherServiceSessionKey;
 
 // Sets the callback queue, specifying that the provided queue is a concurrent queue.
 //
-// When a concurrent queue is explicitly provided via this setter, the the service
-// will wrap the concurrent queue with a new serial queue for each fetcher instance
-// that is created, ensuring all callbacks for a given fetcher instance occur serially.
+// When a concurrent queue is explicitly provided via this setter, then each new fetcher
+// instance created by the service will be provided a new serial queue targeting the
+// concurrent callback queue; this will ensure callbacks for each instance are executed
+// in order, while callbacks from separate fetcher instances are not blocked by each other.
 //
 // The service behavior when resetting the callback queue after providing a concurrent
 // queue is unspecified.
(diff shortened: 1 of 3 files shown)


Source/GTMSessionFetcherService.h
```diff
@@ -107,6 +107,16 @@ extern NSString *const kGTMSessionFetcherServiceSessionKey;
 // fetchers begin.
 - (void)resetSession;
 
+// Sets the callback queue, specifying that the provided queue is a concurrent queue.
+//
+// When a concurrent queue is explicitly provided via this setter, the the service
```
thomasvl, line 112: s/the the/then the/  ?

Source/GTMSessionFetcherService.h
```diff
@@ -107,6 +107,16 @@ extern NSString *const kGTMSessionFetcherServiceSessionKey;
...
+// will wrap the concurrent queue with a new serial queue for each fetcher instance
+// that is created, ensuring all callbacks for a given fetcher instance occur serially.
```
thomasvl, line 114: Continue the comment here to match the commit one, i.e. - "... this allows each fetcher's callbacks to processes concurrently while ensuring an individual fetcher's callbacks are processed serially/in order."

Source/GTMSessionFetcherService.h
mwyman: Done.

Source/GTMSessionFetcherService.h
mwyman: Updated the comment to match the commit message.

thomasvl (approved):

Issue #265, opened 2021-11-24
Static Analysis nullability warning from Xcode 13.1


mwyman:
Xcode 13.1 appears to generate a new static analysis warning:

```
/Source/gtm-session-fetcher/Source/GTMMIMEDocument.m:441:37: warning: nil passed to a callee that requires a non-null 2nd parameter [nullability.NullPassedToNonnull]
        GTMMIMEDocumentPart *part = [GTMMIMEDocumentPart partWithHeaders:headers
                                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

<img width="1133" alt="Screen Shot 2021-11-23 at 4 25 01 PM" src="https://user-images.githubusercontent.com/1154401/143149614-bd010b2a-fc93-49c8-892f-cdfa7bfae67a.png">

PR #266, opened 2021-11-24
Fix static analysis nullability violation finding from Xcode 13.1

mwyman:
Addresses Issue #265.


thomasvl (commented):
Looks like this might trip up one test?

thomasvl (approved):

PR #267, opened 2022-01-18
Log only in verbose mode to reduce the chatter

thomasvl:


Source/GTMSessionFetcher.m
```diff
@@ -2610,8 +2610,8 @@ - (void)invokeFetchCallbacksOnCallbackQueueWithData:(nullable NSData *)data
     }
   }
 
-  GTMSESSION_LOG_DEBUG(@"GTMSessionFetcher invoking fetch callbacks, data %@, error %@", data,
-                       error);
+  GTMSESSION_LOG_DEBUG_VERBOSE(@"GTMSessionFetcher invoking fetch callbacks, data %lu, error %@",
```
mwyman, line 2613: Should this read "data length %lu" rather than "data %lu", just to be clear?

thomasvl: done.

mwyman (approved):

PR #269, opened 2022-02-23
Ignore `-Wunused` warnings around `objc_precise_lifetime` variables.

mwyman:
Several `__attribute__((objc_precise_lifetime))`/`NS_VALID_UNTIL_END_OF_SCOPE`-annotated variables are used to ensure ARC doesn't release variables before we're ready, but the new `-Wunused-but-set-variable` warning introduced in the Xcode 13.3 betas complains about these. Since the variables are not truly _unused_, but have no reason to be accessed beyond their initial setting, it seems better to quash the warning at the specific code rather than annotate the variables `__unused`.

While it wouldn't likely be available in Xcode for quite some time, a patch has been sent to LLVM to exempt `objc_precise_lifetime` variables from the warning: https://reviews.llvm.org/D120372

### DIFF
--- a/Source/GTMMIMEDocument.m
+++ b/Source/GTMMIMEDocument.m
@@ -401,11 +401,11 @@ static void SearchDataForBytes(NSData *data, const void *targetBytes, NSUInteger
         // and map that two-byte subrange.
         const void *partDataBuffer;
         size_t partDataBufferSize;
-        // It's necessary to ignore -Wunused-but-set-variable warnings, as the variables with
-        // objc_precise_lifetime semantics are being used to force ARC to release objects at
-        // a known time.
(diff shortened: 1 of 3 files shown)


thomasvl (commented):
Looks like you might need to `#if` version wrap the `ignored` so it is only seen by clang version that know about it.

mwyman:
> Looks like you might need to `#if` version wrap the `ignored` so it is only seen by clang version that know about it.

Not sure what I'd gate that on, but I can ignore the entire `-Wunused` group around those variables instead, which Clang is happy with.

thomasvl:
> Not sure what I'd gate that on, but I can ignore the entire `-Wunused` group around those variables instead, which Clang is happy with.

You'd likely have to check the CLANG version or something like that.

thomasvl (approved):

PR #278, opened 2022-03-04
Minor fixes found while working on other things

thomasvl:


Source/GTMSessionFetcher.m
```diff
@@ -1916,7 +1916,7 @@ - (void)releaseCallbacks {
   self.retryBlock = nil;
   self.testBlock = nil;
   self.resumeDataBlock = nil;
-  if (@available(iOS 10.0, *)) {
```
mwyman, line 1919: I believe we'd migrated to watchOS 6.0 as a minimum: https://github.com/google/gtm-session-fetcher/pull/239/commits

thomasvl: We don't seem to have it in SwiftPM.  Land this and then update the Package.swift files to cover that afterwards and remove any availability checks?  Or swap to those fixes in this PR?

thomasvl: Updated comment accordingly.

mwyman (approved):

PR #279, opened 2022-03-04
Drop the XML processing/formatting support.

thomasvl:
Was only ever on macOS, and there's always the change the parser/formatter does
something to hide an edge case that was actually in the server's response. This
was most useful back in the GData days since they were all XML based.


mwyman (approved):
Sounds good.

PR #283, opened 2022-03-04
Move chunk generation and provider response handling to a dedicated work queue.

mwyman:
The `chunkWorkQueue` is a serial queue targeting, for now, the default global queue. Creation of chunk data and any file I/O that takes place as a result will now take place on this queue, rather than often being on the main queue.

### DIFF
--- a/Source/GTMSessionUploadFetcher.m
+++ b/Source/GTMSessionUploadFetcher.m
@@ -71,6 +71,9 @@ typedef NS_ENUM(NSUInteger, GTMSessionUploadFetcherStatus) {
 NSString *const kGTMSessionFetcherUploadLocationObtainedNotification =
     @"kGTMSessionFetcherUploadLocationObtainedNotification";
 
+static const char *const kGTMSessionFetcherUploadChunkQueueLabel =
+    "com.google.GTMSessionFetcher.chunkWorkQueue";
+
 #if !GTMSESSION_BUILD_COMBINED_SOURCES
 @interface GTMSessionFetcher (ProtectedMethods)
 
@@ -109,6 +112,8 @@ NSString *const kGTMSessionFetcherUploadLocationObtainedNotification =
 @property(assign, atomic) int64_t uploadGranularity;
 @property(assign, atomic) BOOL allowsCellularAccess;
 
+@property(atomic, readonly) dispatch_queue_t chunkWorkQueue;
+
 @end
 
 @implementation GTMSessionUploadFetcher {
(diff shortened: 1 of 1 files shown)


Source/GTMSessionUploadFetcher.m
```diff
@@ -374,6 +382,27 @@ + (NSArray *)uploadFetchersForBackgroundSessions {
   return uploadFetchers;
 }
 
+- (dispatch_queue_t)chunkWorkQueue {
+  @synchronized(self) {
+    if (!_chunkWorkQueue) {
+      dispatch_queue_t targetQueue = dispatch_get_global_queue(QOS_CLASS_DEFAULT, 0);
+#if TARGET_OS_IOS && (__IPHONE_OS_VERSION_MIN_REQUIRED < __IPHONE_10_0)
```
thomasvl, line 389: I sorta feel like we have this pattern for queue creation a bunch, does it make sense to make a single (inline?) helper for it to reduce the boilerplate until the min iOS version is moved up? (maybe in a follow up PR)

Source/GTMSessionUploadFetcher.m
```diff
@@ -1239,75 +1296,64 @@ - (void)uploadNextChunkWithOffset:(int64_t)offset fetcherProperties:(NSDictionar
...
-                                      }
-                                    }
-                                    if (!didWriteFile) {
-                                      chunkFetcher.bodyData = [chunkData copy];
-                                    }
-                                    [self beginChunkFetcher:chunkFetcher offset:offset];
-                                  });
-                                }];
+
+    GTMSessionUploadFetcherDataProviderResponse response = ^(
```
thomasvl, line 1300: Is this block the same as what was nested below?  (trying to figure out the diffs)

Source/GTMSessionUploadFetcher.m
```diff
@@ -1239,75 +1296,64 @@ - (void)uploadNextChunkWithOffset:(int64_t)offset fetcherProperties:(NSDictionar
...
+        NSData *chunkData, int64_t uploadFileLength, NSError *chunkError) {
```
thomasvl, line 1301: Should this get any queue assertion given the previous comments in the code about contexts?

Source/GTMSessionUploadFetcher.m
mwyman: We've only got one other location at the moment. I was thinking about that, but thought I'd maybe hold off until making a later change to allow setting a worker target queue on the service/fetcher, to allow app subsystems that want to follow Apple's GCD recommendations on a queue hierarchy to set the target for the network activity.

Source/GTMSessionUploadFetcher.m
mwyman: Yes, sorry. `clang-format`'s preferred indentation with the inline block was horrible, and moving it up here made it much cleaner to read.

Source/GTMSessionUploadFetcher.m
mwyman: Done.

thomasvl (approved):

PR #289, opened 2022-03-14
Remove the legacy macros.

thomasvl:


mwyman (approved):
These changes are unrelated to the test failure, which appears to be a race inside a test.

PR #290, opened 2022-03-16
Invoke upload completion handler only after releasing callbacks.

mwyman:
Captures the queue and handler block locally, before releasing all callbacks, and finally invokes the completion handler. This should fix the flake in tests, while being equivalent for overall behavior.

### DIFF
--- a/Source/GTMSessionUploadFetcher.m
+++ b/Source/GTMSessionUploadFetcher.m
@@ -1115,6 +1115,8 @@ static const char *const kGTMSessionFetcherUploadChunkQueueLabel =
 - (void)invokeFinalCallbackWithData:(NSData *)data
                               error:(NSError *)error
            shouldInvalidateLocation:(BOOL)shouldInvalidateLocation {
+  dispatch_queue_t queue;
+  GTMSessionFetcherCompletionHandler handler;
   @synchronized(self) {
     GTMSessionMonitorSynchronized(self);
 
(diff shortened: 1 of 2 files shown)


Source/GTMSessionUploadFetcher.m
```diff
@@ -1115,25 +1115,35 @@ - (BOOL)shouldReleaseCallbacksUponCompletion {
...
-    dispatch_queue_t queue = _delegateCallbackQueue;
-    GTMSessionFetcherCompletionHandler handler = _delegateCompletionHandler;
-    if (queue && handler) {
-      [self invokeOnCallbackQueue:queue
-                 afterUserStopped:NO
-                            block:^{
-                              handler(data, error);
-                            }];
-    }
+    // Wait to dispatch the completion handler until after releasing callbacks. Because
```
thomasvl, line 1127: While all true in how we hit this, does it matter?  or is this change correct from the POV of making the order of the callback and clearing of call backs deterministic?  i.e. - could this trip something else up besides the tests?

mwyman: The prior formulation of this method seems like a relic of the days when the delegate/callback queue was assumed to be the main queue, in which case the callback release was reasonably expected to execute before the callback block could. So this makes the behavior deterministic no matter what queues are being used. I'm inclined to make the same change in `GTMSessionFetcher` as well, but will do that separately.

PR #291, opened 2022-03-22
Use `#import` consistently.

thomasvl:
All other files already use `#import`, so swap these over to be self consistent.

### DIFF
--- a/Source/GTMSessionFetcherLogging.m
+++ b/Source/GTMSessionFetcherLogging.m
@@ -17,8 +17,8 @@
 #error "This file requires ARC support."
 #endif
 
-#include <sys/stat.h>
-#include <unistd.h>
+#import <sys/stat.h>
+#import <unistd.h>
 
 #import "GTMSessionFetcherLogging.h"
 


thomasvl:
The other option is go in the other direction, and change the other places doing `#import` of C system headers to be `#include`.

mwyman:
I think this goes against the ObjC style guide on `#import` vs `#include` being used for [different header types](https://google.github.io/styleguide/objcguide.html#import-and-include) (ObjC/ObjC++ vs C/C++).

Not sure that's so critical for these particular headers, but I remember encountering some that did behave differently between import/include…

thomasvl:
For reference: https://github.com/thomasvl/gtm-session-fetcher/commit/48ca2085f3f58642848f8adbfb81d77e0bd04804 would be going consistent in the other direction.  Let me know if you'd rather this closed and that as a PR.

mwyman:
> For reference: [thomasvl@48ca208](https://github.com/thomasvl/gtm-session-fetcher/commit/48ca2085f3f58642848f8adbfb81d77e0bd04804) would be going consistent in the other direction. Let me know if you'd rather this closed and that as a PR.

Yeah, let's go that direction.

PR #302, opened 2022-04-25
Change the source layout to better match what SwiftPM wants.

thomasvl:
To address the off/on issues with folk trying to use SwiftPM, change the source
layout to be more like what SwiftPM wants.

This also make it so both SwiftPM and CocoaPods users cana refer to all the
headers as `#import <GTMSessionFetcher/HEADER.h>`, this should enable things
that depend on this package to thus also support SwiftPM and CocoaPods more
easily.

- Move sources.
- Update imports as needed.
- Update the files for the packaging systems and the TestApps Xcode project.
- Update github actions to use `files` to directly list things that trigger the
  tests.


thomasvl:
fyi @paulb777 @petea

I'm hoping once this lands and is cut as a release, it will simplify things for GTMAppAuth and Firebase.

mwyman (approved):
File moves and related changes LGTM.

thomasvl:
@karimhm fyi - sorry, forgot to tag you on this when opening the PR.

Issue #314, opened 2022-06-21
`GTMFetcherAuthorizationProtocol` cannot be implemented in Swift


mwyman:
`GTMFetcherAuthorizationProtocol` has a required method, `authorizeRequest:delegate:didFinishSelector:` that cannot be implemented in Swift due to the requirement to use `NSInvocation` to call the provided selector.

The fetcher should update this protocol (or provide a different protocol) providing a more type-safe interface for the authorizer that can be provided by an implementation in Swift.

PR #315, opened 2022-06-28
Deprecating GTMFetcherAuthorizationProtocol in favor of new GTMSessionFetcherAuthorizer

mwyman:
GTMFetcherAuthorizationProtocol uses a id/selector-based callback, which does not play nicely with implementations done in Swift. While the protocol also has an optional block-based form, that block-based callback was never used.

The new protocol implements the old protocol, switching the `@required` and `@optional` methods; meanwhile the old protocol is marked deprecated, along with the `@required` method. This will trigger warnings, which are squashed by conforming to the new protocol and ensuring the block-based form is implemented.

The fetcher is further updated to call the `@optional` method if available, and not call the `@required` authorize-request method unless the `@optional` form does not exist. This allows priority for the block-based form.

Whenever the Fetcher is bumped to v3 the old protocol should be deleted and have its other, non-deprecated methods migrated into the new protocol.

Addresses #314.

### DIFF
--- a/Sources/Core/GTMSessionFetcher.m
+++ b/Sources/Core/GTMSessionFetcher.m
@@ -3814,6 +3814,8 @@ static NSMutableDictionary *gSystemCompletionHandlers = nil;
   }  // @synchronized(self)
 }
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated"
 - (nullable id<GTMFetcherAuthorizationProtocol>)authorizer {
   @synchronized(self) {
     GTMSessionMonitorSynchronized(self);
@@ -3836,6 +3838,7 @@ static NSMutableDictionary *gSystemCompletionHandlers = nil;
     }
   }  // @synchronized(self)
 }
+#pragma clang diagnostic pop
 
 - (nullable NSData *)downloadedData {
   @synchronized(self) {
--- a/Sources/Core/GTMSessionFetcher.m
+++ b/Sources/Core/GTMSessionFetcher.m
@@ -227,7 +227,10 @@ static GTMSessionFetcherTestBlock _Nullable gGlobalTestBlock;
   dispatch_group_t _callbackGroup;   // read-only after creation
   NSOperationQueue *_delegateQueue;  // immutable after beginFetch
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated"
   id<GTMFetcherAuthorizationProtocol> _authorizer;  // immutable after beginFetch
+#pragma clang diagnostic pop
 
   // The service object that created and monitors this fetcher, if any.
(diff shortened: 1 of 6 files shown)


thomasvl (commented):
Looks like the tests need some `#pragma` help also:

```
    In file included from /Users/runner/work/gtm-session-fetcher/gtm-session-fetcher/UnitTests/GTMSessionFetcherFetchingTest.m:24:
    /Users/runner/work/gtm-session-fetcher/gtm-session-fetcher/UnitTests/GTMSessionFetcherFetchingTest.h:74:39: warning: 'GTMFetcherAuthorizationProtocol' is deprecated: implement GTMSessionFetcherAuthorizer instead [-Wdeprecated-declarations]
    @interface TestAuthorizer : NSObject <GTMFetcherAuthorizationProtocol>
```

Sources/Core/GTMSessionFetcher.m
```diff
@@ -1692,14 +1710,14 @@ - (void)authorizeRequest {
   }
 }
 
-- (void)authorizer:(id<GTMFetcherAuthorizationProtocol>)auth
-              request:(NSMutableURLRequest *)authorizedRequest
-    finishedWithError:(NSError *)error {
+- (void)authorizer:(id)auth
```
thomasvl, line 1713: Shouldn't this be nullable also since you use a weak ivar to pass it?

If you do that, you could even make the block always pass nil and just document that we don't use this here any more and it is only there to the selector based signature.

UnitTests/GTMSessionFetcherFetchingTest.m
```diff
@@ -2814,23 +2813,34 @@ - (void)authorizeRequest:(NSMutableURLRequest *)request
...
+           id selfParam = weakSelf;
+           NSMutableURLRequest *requestParam = request;
+           NSMethodSignature *sig = [delegate methodSignatureForSelector:sel];
+           NSInvocation *invocation = [NSInvocation invocationWithMethodSignature:sig];
+           [invocation setSelector:sel];
+           [invocation setTarget:delegate];
+           [invocation setArgument:&selfParam atIndex:2];
+           [invocation setArgument:&requestParam atIndex:3];
+           [invocation setArgument:&error atIndex:4];
+           [invocation setArgument:&error atIndex:4];
```
thomasvl, line 2841: duplicate line

mwyman: Done.

paulb777:
Thanks for this! It made porting the FirebaseStorage Authorizer class to Swift much easier and cleaner. See https://github.com/firebase/firebase-ios-sdk/pull/10117.

What are the plans for including in a release?

thomasvl:
> What are the plans for including in a release?

Thanks for the reminder, 2.1.0 now out.

PR #320, opened 2022-09-13
Fix -wunused-but-set-variable by casting it to void.

itf:
Fix compiler warning about variable badRange being unused but set variables by using idiom of casting it to void.


Sources/Full/GTMMIMEDocument.m
```diff
@@ -315,6 +315,7 @@ + (NSData *)dataWithHeaders:(NSDictionary *)headers {
     NSCharacterSet *badValueChars = [NSCharacterSet characterSetWithCharactersInString:@"\r\n"];
 
     NSRange badRange = [key rangeOfCharacterFromSet:badKeyChars];
+    (void)badRange; //Make compiler detect that badRange is in use to fix warnings
```
mwyman, line 318: I'd rather we just add `__unused` on the badRange variable declaration.

mwyman (changes requested):

thomasvl:
Replaced with #321 321

PR #323, opened 2022-09-29
Update currentOffset from Query Response

maneesht:
Fixes:
* When uploadChunk fails, there is no max retry or exponential backoff until the next request. 
* When a query reports an offset, the next request's offset wasn't updated to reflect that

### DIFF
--- a/Sources/Core/GTMSessionUploadFetcher.m
+++ b/Sources/Core/GTMSessionUploadFetcher.m
@@ -930,11 +930,8 @@ NSString *const kGTMSessionFetcherUploadLocationObtainedNotification =
 - (void)destroyUploadRetryTimer {
   @synchronized(self) {
     GTMSessionMonitorSynchronized(self);
-
-    if (_uploadRetryTimer) {
-      [_uploadRetryTimer invalidate];
-      _uploadRetryTimer = nil;
-    }
+    [_uploadRetryTimer invalidate];
+    _uploadRetryTimer = nil;
   }
 }
 
@@ -943,10 +940,10 @@ NSString *const kGTMSessionFetcherUploadLocationObtainedNotification =
 
   [self setInitialBodyLength:[self bodyLength]];
   if(_minUploadRetryInterval <= 0) {
-    _minUploadRetryInterval = 1;
+    _minUploadRetryInterval = kDefaultMinUploadRetryInterval;
   }
   if(_maxUploadRetryInterval <= 0) {
-    _maxUploadRetryInterval = 60;
+    _maxUploadRetryInterval = kDefaultMaxUploadRetryInterval;
   }
   if(_uploadRetryFactor <= 0) {
     _uploadRetryFactor = 2;
@@ -1971,7 +1968,6 @@ NSString *const kGTMSessionFetcherUploadLocationObtainedNotification =
 // clang-format on
 
 // Internal properties.
-@dynamic uploadRetryTimer;
 @dynamic fetcherInFlight;
 @dynamic activeFetcher;
 @dynamic statusCode;
--- a/Sources/Core/GTMSessionUploadFetcher.m
+++ b/Sources/Core/GTMSessionUploadFetcher.m
@@ -144,9 +144,8 @@ NSString *const kGTMSessionFetcherUploadLocationObtainedNotification =
   BOOL _isPaused;
   BOOL _isRestartedUpload;
   BOOL _shouldInitiateOffsetQuery;
-    
-  NSTimer *_uploadRetryTimer;
 
+  NSTimer *_uploadRetryTimer;
   // Tied to useBackgroundSession property, since this property is applicable to chunk fetchers.
   BOOL _useBackgroundSessionOnChunkFetchers;
 
@@ -939,16 +938,16 @@ NSString *const kGTMSessionFetcherUploadLocationObtainedNotification =
   GTMSessionCheckNotSynchronized(self);
 
   [self setInitialBodyLength:[self bodyLength]];
-  if(_minUploadRetryInterval <= 0) {
+  if (_minUploadRetryInterval <= 0) {
     _minUploadRetryInterval = kDefaultMinUploadRetryInterval;
   }
-  if(_maxUploadRetryInterval <= 0) {
+  if (_maxUploadRetryInterval <= 0) {
     _maxUploadRetryInterval = kDefaultMaxUploadRetryInterval;
   }
-  if(_uploadRetryFactor <= 0) {
+  if (_uploadRetryFactor <= 0) {
     _uploadRetryFactor = 2;
   }
-    
+
   // We'll hold onto the superclass's callback queue so we can invoke the handler
   // even after the superclass has released the queue and its callback handler, as
   // happens during auth failure.
@@ -1405,8 +1404,8 @@ NSString *const kGTMSessionFetcherUploadLocationObtainedNotification =
 
   NSOperationQueue *queue = self.sessionDelegateQueue;
   [queue addOperationWithBlock:^{
-      [self.chunkFetcher beginFetchWithDelegate:self
-                     didFinishSelector:@selector(chunkFetcher:finishedWithData:error:)];
+    [self.chunkFetcher beginFetchWithDelegate:self
+                            didFinishSelector:@selector(chunkFetcher:finishedWithData:error:)];
   }];
 }
 
@@ -1417,8 +1416,6 @@ NSString *const kGTMSessionFetcherUploadLocationObtainedNotification =
     return _uploadRetryTimer;
   }  // @synchronized(self)
 }
-
-
 
 - (NSTimeInterval)maxUploadRetryInterval {
   @synchronized(self) {
@@ -1449,7 +1446,6 @@ NSString *const kGTMSessionFetcherUploadLocationObtainedNotification =
     }
   }  // @synchronized(self)
 }
-
 
 - (NSTimeInterval)minUploadRetryInterval {
   @synchronized(self) {
@@ -1663,7 +1659,8 @@ NSString *const kGTMSessionFetcherUploadLocationObtainedNotification =
   } else {
     int64_t newOffset;
     // The chunk has uploaded successfully.
-    NSString *uploadSizeReceived = [chunkFetcher.responseHeaders objectForKey:kGTMSessionHeaderXGoogUploadSizeReceived];
+    NSString *uploadSizeReceived =
+        [chunkFetcher.responseHeaders objectForKey:kGTMSessionHeaderXGoogUploadSizeReceived];
 
     if (uploadSizeReceived) {
       newOffset = [uploadSizeReceived longLongValue];
--- a/Sources/Core/GTMSessionUploadFetcher.m
+++ b/Sources/Core/GTMSessionUploadFetcher.m
@@ -1376,14 +1376,6 @@ NSString *const kGTMSessionFetcherUploadLocationObtainedNotification =
 
   [self destroyUploadRetryTimer];
 
-#if GTM_BACKGROUND_TASK_FETCHING
-    // Don't keep a background task active while awaiting retry, which can lead to the
-  // app exceeding the allotted time for keeping the background task open, causing the
-  // system to terminate the app. When the retry starts, a new background task will
-  // be created.
-    // I don't think we need to do this here.
-//  [super endBackgroundTask];
-#endif  // GTM_BACKGROUND_TASK_FETCHING
   if (_nextUploadRetryInterval == 0) {
     [self.chunkFetcher beginFetchWithDelegate:self
                             didFinishSelector:@selector(chunkFetcher:finishedWithData:error:)];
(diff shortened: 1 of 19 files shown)


Sources/Core/GTMSessionUploadFetcher.m
```diff
@@ -1513,6 +1513,11 @@ - (void)chunkFetcher:(GTMSessionFetcher *)chunkFetcher
   } else {
     // The chunk has uploaded successfully.
     int64_t newOffset = self.currentOffset + previousContentLength;
+    NSString *uploadSizeReceived = [chunkFetcher.responseHeaders objectForKey:kGTMSessionHeaderXGoogUploadSizeReceived];
```
mwyman, line 1516: Please run clang-format on this (`git clang-format` should do the right thing).

Sources/Core/GTMSessionUploadFetcher.m
```diff
@@ -1513,6 +1513,11 @@ - (void)chunkFetcher:(GTMSessionFetcher *)chunkFetcher
...
+  
+    if(uploadSizeReceived != nil) { // TODO(mtewani): When is this ever nil?
```
mwyman, line 1518: nit: `if (uploadSizeReceived) {`

Sources/Core/GTMSessionUploadFetcher.m
```diff
@@ -1513,6 +1513,11 @@ - (void)chunkFetcher:(GTMSessionFetcher *)chunkFetcher
   } else {
     // The chunk has uploaded successfully.
     int64_t newOffset = self.currentOffset + previousContentLength;
```
mwyman, line 1515: Consider moving the non-uploadSizeReceived case into an `else` clause, since `self.currentOffset` involves a method call and locking. So this could be:

```
int64_t newOffset;
NSString *uploadSizeReceived = ...
if (uploadSizeReceived) {
  newOffset = [uploadSizeReceived longLongValue];
} else {
  newOffset = self.currentOffset + previousContentLength;
}
```

Sources/Core/GTMSessionUploadFetcher.m
```diff
@@ -1755,7 +1760,7 @@ - (int64_t)updateChunkFetcher:(GTMSessionFetcher *)chunkFetcher forChunkAtOffset
   GTMSESSION_ASSERT_DEBUG(offset < fullUploadLength || fullUploadLength == 0,
                           @"offset %lld exceeds data length %lld", offset, fullUploadLength);
 
-  if (granularity > 0) {
+  if (granularity > 0 && offset != fullUploadLength) {
```
mwyman, line 1763: offset < fullUploadLength

? … just for sanity checking?

mwyman (changes requested):

Sources/Core/GTMSessionUploadFetcher.m
maneesht: Done.

Sources/Core/GTMSessionUploadFetcher.m
maneesht: Done.

Sources/Core/GTMSessionUploadFetcher.m
maneesht: Done.

Sources/Core/GTMSessionUploadFetcher.m
maneesht: Done.

UnitTests/GTMSessionFetcherFetchingTest.m
```diff
@@ -3085,10 +3085,11 @@ - (void)fetchStateChanged:(NSNotification *)note {
 
   GTMSessionFetcher *fetcher = note.object;
   BOOL isUploadChunkFetcher = ([fetcher parentUploadFetcher] != nil);
-  BOOL isFetchStartedNotification = [note.name isEqual:kGTMSessionFetcherStartedNotification];
+    BOOL isFetchStartedNotification = [note.name isEqual:kGTMSessionFetcherStartedNotification];
 
   if (isFetchStartedNotification) {
     ++_fetchStarted;
+      NSLog(@"%d", _fetchStarted);
```
mwyman, line 3092: `GTMSESSION_LOG_DEBUG`

UnitTests/GTMSessionFetcherChunkedUploadTest.m
```diff
@@ -809,6 +809,201 @@ - (void)testBigFileURLSingleChunkedUploadFetchRetry {
...
+
+  BOOL (^shouldRetryUpload)(GTMSessionUploadFetcher *, BOOL, NSError *) =
+      ^BOOL(GTMSessionUploadFetcher *blockFetcher, BOOL suggestedWillRetry, NSError *error) {
+        // Change this fetch's request to have the original, non-failure status URL.
+        // This will make the retry succeed.
+        return suggestedWillRetry;  // do the retry fetch; it should succeed now
+      };
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Warc-retain-cycles"
```
mwyman, line 843: Can we capture the fetcher weakly to avoid the `-Warc-retain-cycles` warning? I would prefer we avoid introducing retain cycles in the tests.

UnitTests/GTMSessionFetcherChunkedUploadTest.m
```diff
@@ -809,6 +809,201 @@ - (void)testBigFileURLSingleChunkedUploadFetchRetry {
   [self removeTemporaryFileURL:bigFileURL];
 }
 
+- (void)testBigFileURLSingleChunkedUploadFetchRetryWithOffset {
```
mwyman, line 812: Can we also add a test to trigger cancellation of the upload while it's waiting on a retry?

mwyman (changes requested):

mwyman (changes requested):

Sources/Core/GTMSessionUploadFetcher.m
```diff
@@ -1320,6 +1374,123 @@ - (void)uploadNextChunkWithOffset:(int64_t)offset fetcherProperties:(NSDictionar
   }
 }
 
+- (void)beginRetryTimer {
```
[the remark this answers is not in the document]
maneesht: Absolutely. Meant to be a part of my next round of cleanup. Thanks

Sources/Core/Public/GTMSessionFetcher/GTMSessionUploadFetcher.h
```diff
@@ -133,6 +133,15 @@ typedef void (^GTMSessionUploadFetcherCancellationHandler)(GTMSessionFetcher *_N
 @property(atomic, copy) NSString *uploadMIMEType;
 @property(atomic, readonly, assign) int64_t chunkSize;
 @property(atomic, readonly, assign) int64_t currentOffset;
+@property(atomic, assign) double uploadRetryFactor;
+@property(atomic, assign)  NSTimeInterval nextUploadRetryInterval;
```
mwyman, line 137: Why does `nextUploadRetryInterval` need to be public? Could this just be an ivar? Or should we follow the fetcher behavior and store `lastUploadRetryInterval` and just calculate the next one via a method when it's needed?

UnitTests/GTMSessionFetcherChunkedUploadTest.m
```diff
@@ -809,6 +809,130 @@ - (void)testBigFileURLSingleChunkedUploadFetchRetry {
...
+  fetcher.cancellationHandler = ^(GTMSessionFetcher *_Nullable sessionFetcher,
+                                  NSData *_Nullable data, NSError *_Nullable error) {
+    __strong typeof(fetcher) strongFetcher = weakFetcher;
+    XCTAssertNil(strongFetcher.uploadRetryTimer);
+    XCTAssertNotNil(sessionFetcher);
+    [expectation fulfill];
+  };
+  fetcher.backoffBlock = ^() {
+    __strong typeof(fetcher) strongFetcher = weakFetcher;
+    [strongFetcher performSelector:@selector(stopFetching) withObject:nil afterDelay:0.0];
```
mwyman, line 901: To call a single method, the weak->strong conversion here doesn't need to be explicit:

```
[weakFetcher performSelector:@selector(stopFetching) withObject:nil afterDelay:0.0];
```

mwyman (changes requested):
And run clang-format, please :)

Sources/Core/Public/GTMSessionFetcher/GTMSessionUploadFetcher.h
maneesht: Lets just store the last one. For some reason I thought GTMSessionFetcher had stored the next one.

Sources/Core/GTMSessionUploadFetcher.m
```diff
@@ -1320,6 +1363,114 @@ - (void)uploadNextChunkWithOffset:(int64_t)offset fetcherProperties:(NSDictionar
...
+}
+}
+
+- (void)uploadRetryTimerFired:(NSTimer *)timer {
+  [self destroyUploadRetryTimer];
+
+  NSOperationQueue *queue = self.sessionDelegateQueue;
+  [queue addOperationWithBlock:^{
+      [self.chunkFetcher beginFetchWithDelegate:self
+                     didFinishSelector:@selector(chunkFetcher:finishedWithData:error:)];
```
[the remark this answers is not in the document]
maneesht: Re-ran

thomasvl:
You may also need to merge in the changes on main (or rebase) so the CI tests fully pass.

Sources/Core/GTMSessionUploadFetcher.m
```diff
@@ -905,11 +927,31 @@ - (void)beginFetchForRetry {
   [super beginFetchForRetry];
 }
 
+- (void)destroyUploadRetryTimer {
+  @synchronized(self) {
+    GTMSessionMonitorSynchronized(self);
+
+    if (_uploadRetryTimer) {
```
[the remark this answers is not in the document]
thomasvl: > Will remove. Why isn't it needed?

ObjC Supports messaging `nil` (becomes a noop).

> It looks like we're checking in `GTMSessionFetcher`

Other case we aren't, guess we aren't consistent. 😢

Sources/Core/Public/GTMSessionFetcher/GTMSessionUploadFetcher.h
```diff
@@ -133,6 +133,14 @@ typedef void (^GTMSessionUploadFetcherCancellationHandler)(GTMSessionFetcher *_N
 @property(atomic, copy) NSString *uploadMIMEType;
 @property(atomic, readonly, assign) int64_t chunkSize;
 @property(atomic, readonly, assign) int64_t currentOffset;
+@property(atomic, assign) double uploadRetryFactor;
+@property(atomic, assign)  NSTimeInterval maxUploadRetryInterval;
+@property(atomic, assign)  NSTimeInterval minUploadRetryInterval;
+@property(atomic, readonly) NSTimer *uploadRetryTimer;
```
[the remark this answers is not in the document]
thomasvl: I'd have to double check, but I think we already declare some stuff like this in the impl and tests so the test can use it without promoting it to fully public interface.  I'd rather not make this public interface if there is no reason a consumer should need it.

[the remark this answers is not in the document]
maneesht: Actually, I forgot I added a getter in the implementation, so I'll get rid of this in the next commit.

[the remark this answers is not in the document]
thomasvl: Looks like it is still in the header.

Sources/Core/Public/GTMSessionFetcher/GTMSessionUploadFetcher.h
```diff
@@ -133,6 +133,13 @@ typedef void (^GTMSessionUploadFetcherCancellationHandler)(GTMSessionFetcher *_N
 @property(atomic, copy) NSString *uploadMIMEType;
 @property(atomic, readonly, assign) int64_t chunkSize;
 @property(atomic, readonly, assign) int64_t currentOffset;
+@property(atomic, assign) double uploadRetryFactor;
+@property(atomic, assign) NSTimeInterval maxUploadRetryInterval;
+@property(atomic, assign) NSTimeInterval minUploadRetryInterval;
+typedef void (^GTMSessionFetcherExponentialBackoffBlock)(void);
```
thomasvl, line 139: Actually, rereading things - how is a developer supposed to use this backoff block?  i.e. - what would they need it for? The test is using it, but that usage doesn't seem like what it would be generally used for.

thomasvl: If this is just for the test, can we use the test hook or maybe the retry notifications in some way to hook in and cancel the fetcher instead?

thomasvl (approved):
Thanks!

PR #324, opened 2022-10-05
Replace `dispatch_semaphore` use with `os_unfair_lock`.

mwyman:
Xcode 14 thread checker has flagged potential priority inversion issues relating to use of `dispatch_semaphore` when handling session creation and dispatcher resetting. This replaces uses of semaphores with `os_unfair_lock` which is available on iOS 10+ and all other platforms we're concerned with.

To avoid the danger of leaking the lock, this also refactors session creation by having `GTMSessionFetcher` instances provide a session creation block to the service, which is only called if the service does not have an active `NSURLSession`.

### DIFF
--- a/Sources/Core/GTMSessionFetcherService.m
+++ b/Sources/Core/GTMSessionFetcherService.m
@@ -290,28 +290,31 @@ NSString *const kGTMSessionFetcherServiceSessionKey = @"kGTMSessionFetcherServic
     }
   }
 
-  NSURLSession *session;
+  @try {
+    NSURLSession *session;
+    // Wait if another fetcher is currently creating a session; avoid waiting inside the
+    // @synchronized block as that could deadlock.
+    os_unfair_lock_lock(&_sessionCreationLock);
+    @synchronized(self) {
+      GTMSessionMonitorSynchronized(self);
 
-  // Wait if another fetcher is currently creating a session; avoid waiting inside the
-  // @synchronized block as that could deadlock.
-  os_unfair_lock_lock(&_sessionCreationLock);
-  @synchronized(self) {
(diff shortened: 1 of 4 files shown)


Sources/Core/GTMSessionFetcherService.m
```diff
@@ -315,18 +316,14 @@ - (NSURLSession *)sessionForFetcherCreation {
...
-      // The calling fetcher will receive a preexisting session, so
-      // we can allow other fetchers to create a session.
-      dispatch_semaphore_signal(_sessionCreationSemaphore);
-    } else {
-      // No existing session was obtained, so the calling fetcher will create the session;
-      // it *must* invoke fetcherDidCreateSession: to signal the dispatcher's semaphore after
-      // the session has been created (or fails to be created) to avoid a hang.
+    session = _delegateDispatcher.session;
+    if (!session) {
+      session = creationBlock(_delegateDispatcher);
```
thomasvl, line 321: Try/catch on this to avoid going to a bad state?

thomasvl:
Do we also raise the min versions?

thomasvl:
> Do we also raise the min versions?

I thought I saw one build error about availability, but now I don't see it.

Sources/Core/GTMSessionFetcherService.m
thomasvl: Or maybe just a try/finally?  Or are you comfortable in it should never throw?

mwyman: Was thinking it should be fine, but for safety's sake yeah it should probably have try/finally to ensure the lock is released.

thomasvl (approved):

thomasvl (approved):

PR #327, opened 2022-10-07
Bump minimum iOS version to 10.0.

mwyman:
This will allow use of some iOS 10-minimum features, and any other cleanup.

Firebase updated iOS minimum (https://github.com/firebase/firebase-ios-sdk/pull/6517).

### DIFF
--- a/.github/workflows/cocoapods.yml
+++ b/.github/workflows/cocoapods.yml
@@ -28,7 +28,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        PLATFORM: ["ios", "macos", "tvos", "watchos"]
+        # Add back watchOS. See CocoaPods/CocoaPods#11558
+        PLATFORM: ["ios", "macos", "tvos"]
         CONFIGURATION: ["Debug", "Release"]
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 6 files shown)


thomasvl (dismissed):

paulb777:
FYI on the watchos pod-lib-lint failures: https://github.com/firebase/firebase-ios-sdk/issues/10316

thomasvl:
> FYI on the watchos pod-lib-lint failures: [firebase/firebase-ios-sdk#10316](https://github.com/firebase/firebase-ios-sdk/issues/10316)

Thanks, what do you do for pushes then since I'd assume this will show up for those also?

paulb777:
We haven't tried to push yet, but we'll likely use Xcode 13 like others in the CocoaPods issue thread.

`--skip-import-validation` might be a workaround, but I haven't tried.

thomasvl (dismissed):

thomasvl (approved):

jokerduke:
Since the bump to iOS 10, I get the error with GTMFetcher on GoogelSinIn and GTMAppAuth. The error is:

The package product 'GTMSessionFetcherCore' requires minimum platform version 10.0 for the iOS platform, but this target supports 9.0

GTMAuth has target iOS v9.0 on the package.swift as GoogleSignIn.

Any suggestions on how  I can solve this?

Thanks

thomasvl:
We haven't made a release yet with this, so I'm sorta surprised. I guess you aren't using releases?

Since Firebase moved up their minimums, we expect these others will eventually do so also.

PR #328, opened 2022-10-10
Add API to call callbacks after a cancel

### LINKED ISSUE #318: Improve async/await support
If a fetch is in progress, via a `CheckedCancellation` wrapper on `beginFetch`, and `stopFetching` is called, the following hang occurs:
```
SWIFT TASK CONTINUATION MISUSE: beginFetch(uploadFetcher:) leaked its continuation!
2022-09-02 13:47:25.949984-0700 AppHost-FirebaseStorage-Unit-Tests[38388:17164810] SWIFT TASK CONTINUATION MISUSE: beginFetch(uploadFetcher:) leaked its continuation! 
```

From:
```
  private func beginFetch(uploadFetcher: GTMSessionUploadFetcher) async -> (Data?, Error?) {
    return await withCheckedContinuation { continuation in
      uploadFetcher.beginFetch { data, error in
        continuation.resume(returning: (data, error))
      }
    }
  }
```

I suspect that the right solution might be to add an additional stopFetching API variation that calls back its completion handlers instead of deallocating them.

More context at https://github.com/firebase/firebase-ios-sdk/pull/10161 and internally at https://groups.google.com/a/google.com/g/gtm-session-fetcher-eng/c/AFt3g1V_tsw/m/_gIYdK7bAAAJ?utm_medium=email&utm_source=footer

paulb777:
Fix #318

### DIFF
--- a/Sources/Core/GTMSessionFetcher.m
+++ b/Sources/Core/GTMSessionFetcher.m
@@ -1996,7 +1996,7 @@ NSData *_Nullable GTMDataFromInputStream(NSInputStream *inputStream, NSError **o
   [holdSelf destroyRetryTimer];
 
   BOOL sendStopNotification = YES;
-  BOOL shouldStopFetcher = YES;
+  BOOL cancelStopFetcher = NO;
   @synchronized(self) {
     GTMSessionMonitorSynchronized(self);
 
@@ -2068,7 +2068,7 @@ NSData *_Nullable GTMDataFromInputStream(NSInputStream *inputStream, NSError **o
         }
       }
     }
-    shouldStopFetcher = !_stopFetchingTriggersCompletionHandler || !_userStoppedFetching;
+    cancelStopFetcher = _stopFetchingTriggersCompletionHandler && _userStoppedFetching;
   }  // @synchronized(self)
 
   // If the NSURLSession needs to be invalidated, but needs to wait until the delegate method
@@ -2086,7 +2086,7 @@ NSData *_Nullable GTMDataFromInputStream(NSInputStream *inputStream, NSError **o
     self.authorizer = nil;
   }
 
-  if (shouldStopFetcher) {
+  if (!cancelStopFetcher) {
     [service fetcherDidStop:self];
   }
 
--- a/Sources/Core/GTMSessionFetcher.m
+++ b/Sources/Core/GTMSessionFetcher.m
@@ -1995,6 +1995,7 @@ NSData *_Nullable GTMDataFromInputStream(NSInputStream *inputStream, NSError **o
   [holdSelf destroyRetryTimer];
 
   BOOL sendStopNotification = YES;
+  BOOL shouldStopFetcher = YES;
   @synchronized(self) {
     GTMSessionMonitorSynchronized(self);
 
@@ -2066,6 +2067,7 @@ NSData *_Nullable GTMDataFromInputStream(NSInputStream *inputStream, NSError **o
         }
       }
     }
+    shouldStopFetcher = !_stopFetchingTriggersCompletionHandler || !_userStoppedFetching;
   }  // @synchronized(self)
 
   // If the NSURLSession needs to be invalidated, but needs to wait until the delegate method
@@ -2083,7 +2085,7 @@ NSData *_Nullable GTMDataFromInputStream(NSInputStream *inputStream, NSError **o
     self.authorizer = nil;
   }
 
-  if (!self.stopFetchingTriggersCompletionHandler) {
+  if (shouldStopFetcher) {
     [service fetcherDidStop:self];
   }
 
--- a/Sources/Core/GTMSessionFetcher.m
+++ b/Sources/Core/GTMSessionFetcher.m
@@ -4020,11 +4020,13 @@ static NSMutableDictionary *gSystemCompletionHandlers = nil;
 }
 
 - (void)setStopFetchingTriggersCompletionHandler:(BOOL)flag {
-  GTMSESSION_ASSERT_DEBUG(
-      _initialBeginFetchDate == nil,
-      @"stopFetchingTriggersCompletionHandler should not change after fetcher starts");
-
-  _stopFetchingTriggersCompletionHandler = flag;
+  if (_initialBeginFetchDate == nil) {
+    _stopFetchingTriggersCompletionHandler = flag;
+  } else {
+    GTMSESSION_ASSERT_DEBUG(
+        0,
+        @"stopFetchingTriggersCompletionHandler should not change after fetcher starts");
+  }
 }
 
 - (nullable NSURLSession *)sessionNeedingInvalidation {
--- a/Sources/Core/GTMSessionFetcher.m
+++ b/Sources/Core/GTMSessionFetcher.m
@@ -142,7 +142,6 @@ NS_ASSUME_NONNULL_END
 
 @property(atomic, strong, readwrite, nullable) NSData *downloadedData;
 @property(atomic, strong, readwrite, nullable) NSData *downloadResumeData;
-@property(atomic, readwrite) BOOL cancelAndCallback;
 
 #if GTM_BACKGROUND_TASK_FETCHING
 // Should always be accessed within an @synchronized(self).
@@ -214,7 +213,6 @@ static GTMSessionFetcherTestBlock _Nullable gGlobalTestBlock;
   NSString *_sessionIdentifierUUID;
   BOOL _userRequestedBackgroundSession;
   BOOL _usingBackgroundSession;
-  BOOL _cancelAndCallback;
   NSMutableData *_Nullable _downloadedData;
   NSError *_downloadFinishedError;
   NSData *_downloadResumeData;               // immutable after construction
@@ -1966,27 +1964,27 @@ NSData *_Nullable GTMDataFromInputStream(NSInputStream *inputStream, NSError **o
   }
 }
 
-// External stop method and don't call callbacks
+// External stop method
 - (void)stopFetching {
-  [self stopFetching:NO];
-}
-
-// External stop method with do call callbacks option
-- (void)stopFetching:(BOOL)shouldCallCallbacks {
   @synchronized(self) {
     GTMSessionMonitorSynchronized(self);
 
     // Prevent enqueued callbacks from executing.
     _userStoppedFetching = YES;
   }  // @synchronized(self)
-  if (shouldCallCallbacks) {
+
+  if (self.stopFetchingTriggersCallbacks) {
     NSError *error = [NSError errorWithDomain:kGTMSessionFetcherErrorDomain
-                                   code:GTMSessionFetcherErrorCancelled
+                                         code:GTMSessionFetcherErrorCancelled
                                      userInfo:@{@"description" : @"Operation cancelled"}];
-    _cancelAndCallback = YES;
-    _completionHandler(nil, error);
+    [self invokeFetchCallbacksOnCallbackQueueWithData:nil
+                                                error:error
+                                          mayDecorate:YES
+                               shouldReleaseCallbacks:YES];
+    [self sendStopNotificationIfNeeded];
+  } else {
+    [self stopFetchReleasingCallbacks:YES];
   }
-  [self stopFetchReleasingCallbacks:YES];
 }
 
 // Cancel the fetch of the URL that's currently in progress.
@@ -2614,7 +2612,7 @@ static _Nullable id<GTMUIApplicationProtocol> gSubstituteUIApp;
                         block:(void (^)(void))block {
   if (callbackQueue) {
     dispatch_group_async(_callbackGroup, callbackQueue, ^{
-      if (!afterStopped && !self->_cancelAndCallback) {
+      if (!afterStopped && !self->_stopFetchingTriggersCallbacks) {
         NSDate *serviceStoppedAllDate = [self->_service stoppedAllFetchersDate];
 
         @synchronized(self) {
@@ -3683,7 +3681,7 @@ static NSMutableDictionary *gSystemCompletionHandlers = nil;
             testBlockAccumulateDataChunkCount = _testBlockAccumulateDataChunkCount,
             comment = _comment,
             log = _log,
-            cancelAndCallback = _cancelAndCallback;
+            stopFetchingTriggersCallbacks = _stopFetchingTriggersCallbacks;
 
 #if !STRIP_GTM_FETCH_LOGGING
 @synthesize redirectedFromURL = _redirectedFromURL,
@@ -4007,6 +4005,22 @@ static NSMutableDictionary *gSystemCompletionHandlers = nil;
     GTMSessionMonitorSynchronized(self);
 
     _usingBackgroundSession = flag;
+  }  // @synchronized(self)
+}
+
+- (BOOL)stopFetchingTriggersCallbacks {
+  @synchronized(self) {
+    GTMSessionMonitorSynchronized(self);
+
+    return _stopFetchingTriggersCallbacks;
+  }  // @synchronized(self)
+}
+
+- (void)setStopFetchingTriggersCallbacks:(BOOL)flag {
+  @synchronized(self) {
+    GTMSessionMonitorSynchronized(self);
+
+    _stopFetchingTriggersCallbacks = flag;
   }  // @synchronized(self)
 }
 
--- a/Sources/Core/GTMSessionFetcher.m
+++ b/Sources/Core/GTMSessionFetcher.m
@@ -1972,7 +1972,6 @@ NSData *_Nullable GTMDataFromInputStream(NSInputStream *inputStream, NSError **o
     // Prevent enqueued callbacks from executing.
     _userStoppedFetching = YES;
   }  // @synchronized(self)
-
   if (self.stopFetchingTriggersCallbacks) {
     NSError *error = [NSError errorWithDomain:kGTMSessionFetcherErrorDomain
                                          code:GTMSessionFetcherErrorCancelled
@@ -1982,9 +1981,9 @@ NSData *_Nullable GTMDataFromInputStream(NSInputStream *inputStream, NSError **o
                                           mayDecorate:YES
                                shouldReleaseCallbacks:YES];
     [self sendStopNotificationIfNeeded];
-  } else {
-    [self stopFetchReleasingCallbacks:YES];
   }
+  [self stopFetchReleasingCallbacks:!self.stopFetchingTriggersCallbacks];
+
 }
 
 // Cancel the fetch of the URL that's currently in progress.
--- a/Sources/Core/GTMSessionFetcher.m
+++ b/Sources/Core/GTMSessionFetcher.m
@@ -4020,16 +4020,15 @@ static NSMutableDictionary *gSystemCompletionHandlers = nil;
 }
 
 - (BOOL)stopFetchingTriggersCompletionHandler {
-  @synchronized(self) {
-    GTMSessionMonitorSynchronized(self);
-
-    return _stopFetchingTriggersCompletionHandler;
-  }  // @synchronized(self)
+  return _stopFetchingTriggersCompletionHandler;
 }
 
 - (void)setStopFetchingTriggersCompletionHandler:(BOOL)flag {
   @synchronized(self) {
     GTMSessionMonitorSynchronized(self);
+    GTMSESSION_ASSERT_DEBUG(
+        _initialBeginFetchDate == nil,
+        @"stopFetchingTriggersCompletionHandler should not change after fetcher starts");
 
     _stopFetchingTriggersCompletionHandler = flag;
   }  // @synchronized(self)
--- a/Sources/Core/GTMSessionFetcher.m
+++ b/Sources/Core/GTMSessionFetcher.m
@@ -1969,7 +1969,8 @@ NSData *_Nullable GTMDataFromInputStream(NSInputStream *inputStream, NSError **o
   @synchronized(self) {
     GTMSessionMonitorSynchronized(self);
 
-    // Prevent enqueued callbacks from executing.
+    // Prevent enqueued callbacks from executing. The completion handler will still execute if
+    // the property `stopFetchingTriggersCompletionHandler` is `YES`.
     _userStoppedFetching = YES;
   }  // @synchronized(self)
   [self stopFetchReleasingCallbacks:!self.stopFetchingTriggersCompletionHandler];
@@ -2605,8 +2606,11 @@ static _Nullable id<GTMUIApplicationProtocol> gSubstituteUIApp;
 
   if (callbackQueue) {
     dispatch_group_async(_callbackGroup, callbackQueue, ^{
-      if (!afterStopped &&
-          (!self->_stopFetchingTriggersCompletionHandler || !self->_userStoppedFetching)) {
+      BOOL skipEarlyExit = NO;
+      @synchronized(self) {
+        skipEarlyExit = !self->_stopFetchingTriggersCompletionHandler;
+      }
+      if (!afterStopped && skipEarlyExit) {
         NSDate *serviceStoppedAllDate = [self->_service stoppedAllFetchersDate];
 
         @synchronized(self) {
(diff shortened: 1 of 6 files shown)


Sources/Core/GTMSessionFetcher.m
```diff
@@ -139,6 +139,7 @@ @interface GTMSessionFetcher ()
 
 @property(atomic, strong, readwrite, nullable) NSData *downloadedData;
 @property(atomic, strong, readwrite, nullable) NSData *downloadResumeData;
+@property(atomic, readwrite) BOOL cancelAndCallback;
```
mwyman, line 142: Note: all the fetcher's `atomic` properties are explicitly implemented and do `@synchronized(self)` rather than rely on the generated "atomic" getter/setter (which almost never does what devs really want).

Sources/Core/Public/GTMSessionFetcher/GTMSessionFetcher.h
```diff
@@ -1103,6 +1104,10 @@ __deprecated_msg("implement GTMSessionFetcherAuthorizer instead")
 // will not be called.
 - (void)stopFetching;
 
+// Cancel the fetch of the request that's currently in progress.  The completion handler
+// is called on option.
+- (void)stopFetching:(BOOL)shouldCallCallbacks;
```
mwyman, line 1109: My understanding is that this change is being driven by desire for Swift `async` support, and if Swift code demands the continuation is always called then the API should make sure than an appropriately Swift-configured fetcher will always trigger callbacks on `stopFetching` being called, without the code canceling the fetch needing to know to call a different `stopFetching` variant.

I think this needs to be reconsidered and made into a single fire-and-forget kind of approach, where Swift code can set a `fetcher.stopFetchingTriggersCallbacks = true` kind of property, at which point any call to `stopFetching` will trigger the callbacks with the appropriate error.

UnitTests/GTMSessionFetcherFetchingTest.m
```diff
@@ -1720,6 +1720,40 @@ - (void)testQuickBeginStopFetching {
 #endif
 }
 
+- (void)testQuickBeginDontStopFetching {
```
mwyman, line 1723: This seems unrelated to the PR at hand; can we simply handle the completion callback-with-error behavior, and assert that cancellation triggers an appropriate error in the fetcher?

Consider having your test access the local HTTP `_testServer` with a "sleep=" query parameter so you can have time to cancel the fetch on the client side.

Sources/Core/GTMSessionFetcher.m
```diff
@@ -1964,14 +1966,26 @@ - (void)forgetSessionIdentifierForFetcherWithoutSyncCheck {
...
     GTMSessionMonitorSynchronized(self);
 
     // Prevent enqueued callbacks from executing.
     _userStoppedFetching = YES;
   }  // @synchronized(self)
+  if (shouldCallCallbacks) {
+    NSError *error = [NSError errorWithDomain:kGTMSessionFetcherErrorDomain
+                                   code:GTMSessionFetcherErrorCancelled
+                                     userInfo:@{@"description" : @"Operation cancelled"}];
+    _cancelAndCallback = YES;
```
mwyman, line 1986: Note that access to `atomic` properties underlying ivars are not implicitly synchronized (ever), and so either need to use `.` property access or be wrapped in the appropriate locking (if the getter/setter are implemented manually).

Sources/Core/GTMSessionFetcher.m
```diff
@@ -1964,14 +1966,26 @@ - (void)forgetSessionIdentifierForFetcherWithoutSyncCheck {
...
+    _completionHandler(nil, error);
```
mwyman, line 1987: The completion handler block should not be invoked directly from the thread calling `stopFetching`, among other things because the completion handler block must be invoked on the `callbackQueue`.

Also, access to the `_completionHandler` ivar needs to be guarded by `@synchronized(self)`.

It would probably be best to add parameter(s) to `stopFetchingReleasingCallbacks:` that triggers invoking the completion handler prior to calling `releaseCallbacks`, following the behavior pattern of `invokeFetchCallbacksOnCallbackQueueWithData:error:mayDecorate:shouldReleaseCallbacks:`, or possibly calls that method directly.

mwyman (changes requested):

UnitTests/GTMSessionFetcherFetchingTest.m
paulb777: This test directly contrasts with the test before it to verify that the callbacks do get called, where they didn't in the previous test. Isn't that useful?

paulb777 (commented):
Thanks for the feedback. After being out much of the last two weeks, I'm finally back at it ...

UnitTests/GTMSessionFetcherFetchingTest.m
mwyman: The test before this is testing a specific workaround; testing that the callback is triggered on cancel requires different behavior, which I described in my original comment.

mwyman (changes requested):

Sources/Core/GTMSessionFetcher.m
```diff
@@ -1972,7 +1972,19 @@ - (void)stopFetching {
...
+
+  if (self.stopFetchingTriggersCallbacks) {
+    NSError *error = [NSError errorWithDomain:kGTMSessionFetcherErrorDomain
+                                         code:GTMSessionFetcherErrorCancelled
+                                     userInfo:@{@"description" : @"Operation cancelled"}];
+    [self invokeFetchCallbacksOnCallbackQueueWithData:nil
+                                                error:error
+                                          mayDecorate:YES
+                               shouldReleaseCallbacks:YES];
+    [self sendStopNotificationIfNeeded];
```
mwyman, line 1984: This doesn't trigger stopping the fetch, just invokes the callback and sends the notification.

I think the changes needs to be spread across `stopFetchingReleasingCallbacks:` and `URLSession:task:didCompleteWithError:`.

Here, instead of this code we pass `NO` if `self.stopFetchingTriggersCallbacks` to this initial `stopFetchingReleasingCallbacks:` and in `URLSession:task:didCompleteWithError:` checks for an error with `NSURLErrorDomain` and code `NSURLErrorCancelled` and perhaps wraps that error before calling the callback.

WDYT?

mwyman (changes requested):

Sources/Core/GTMSessionFetcher.m
paulb777: I couldn't get the changes to work in `stopFetchingReleasingCallbacks` because of other call paths into it.   Instead, I reverted to always calling  `stopFetchingReleasingCallbacks`  from `stopFetching`. 

I'm also not sure what to do in `URLSession:task:didCompleteWithError:`. Based on https://github.com/google/gtm-session-fetcher/blob/main/UnitTests/GTMSessionFetcherFetchingTest.m#L886, it looks like callbacks occur if the cancellation happens on the server.

Sources/Core/GTMSessionFetcher.m
```diff
@@ -2600,7 +2611,7 @@ - (void)invokeOnCallbackQueue:(dispatch_queue_t)callbackQueue
                         block:(void (^)(void))block {
   if (callbackQueue) {
     dispatch_group_async(_callbackGroup, callbackQueue, ^{
-      if (!afterStopped) {
+      if (!afterStopped && !self->_stopFetchingTriggersCallbacks) {
```
mwyman, line 2614: I worry this will let other blocks slip through if `_stopFetchingTriggersCallbacks` is true, and that may not be what we want.

For Swift `async` my understanding is the only callback we want firing after `stopFetching` is called is the completion handler because that's the one that needs to trigger the continuation. Are we sure we want to allow other callback blocks to be sent? Otherwise, I think the `afterStopped` should be what we adjust if `.stopFetchingTriggersCallbacks` is set.

Sources/Core/Public/GTMSessionFetcher/GTMSessionFetcher.h
```diff
@@ -1100,9 +1101,12 @@ __deprecated_msg("implement GTMSessionFetcherAuthorizer instead")
 @property(atomic, readonly, getter=isFetching) BOOL fetching;
 
 // Cancel the fetch of the request that's currently in progress.  The completion handler
-// will not be called.
+// will be called if the property `stopFetchingTriggersCallbacks` is `YES`.
 - (void)stopFetching;
 
+// Call callbacks in `stopFetching`.
+@property(atomic, assign) BOOL stopFetchingTriggersCallbacks;
```
mwyman, line 1108: Should this be limited to the completion handler?

`stopFetchingTriggersCompletionHandler` ?

Sources/Core/GTMSessionFetcher.m
mwyman: My worry is that this is a new codepath to firing the completion handler, and I'm not sure the implications of firing the handler here before the rest of the flow has completed.

My suggestion is that we allow the task cancellation to take place, but in `finishWithError:shouldRetry:` we would want to check for a cancellation error from NSURLSession, and if so check whether `_userStoppedFetching` is set, and let the callback flow happen that way. Would also need to ensure the completion handler block will fire after user-stopped for that case.

When `cancel` is called on a task, `NSURLErrorCancelled` is returned: https://developer.apple.com/documentation/foundation/nsurlsessiontask/1411591-cancel

mwyman (changes requested):

paulb777:
Moving to `finishWithError:shouldRetry:` seems to work except for the case where the cancel happens after a successful upload but before the success completion is called. I can't find where the continuation gets leaked in that case.

paulb777:
Mostly addressed the comments. Open issues are three test failures because of fetcher.completionHander not being nilled out and correctly passing the `afterStopped` parameter

Sources/Core/GTMSessionFetcher.m
```diff
@@ -2991,6 +3000,14 @@ - (void)URLSession:(NSURLSession *)session
     }
   }  // @synchronized(self)
 
+  if (self.stopFetchingTriggersCallbacks && self.userStoppedFetching) {
+    NSError *error = [NSError errorWithDomain:kGTMSessionFetcherErrorDomain
+                                         code:GTMSessionFetcherErrorCancelled
+                                     userInfo:@{@"description" : @"Operation cancelled"}];
```
mwyman, line 3006: @"description" -> NSLocalizedDescriptionKey

mwyman, line 3006: Would we want to also add NSUnderlyingErrorKey with the cancellation error from NSURLSession? Not sure whether I think so, but wanted to ask the question.

UnitTests/GTMSessionFetcherChunkedUploadTest.m
```diff
@@ -1537,6 +1577,63 @@ - (void)testBigDataChunkedUploadWithCancel {
...
+}
+
+- (void)testBigDataChunkedUploadWithCancelAndCallback {
+  // Repeat the previous test, canceling after 20,000 bytes.
+  CREATE_START_STOP_NOTIFICATION_EXPECTATIONS(3, 3);
+  FetcherNotificationsCounter *fnctr = [[FetcherNotificationsCounter alloc] init];
+
+  NSURLRequest *request = [self validUploadFileRequest];
+  NSData *bigData = [self bigUploadData];
+  GTMSessionUploadFetcher *fetcher = [GTMSessionUploadFetcher uploadFetcherWithRequest:request
```
mwyman, line 1600: Mind making sure you've run `git clang-format` on this PR?

mwyman (changes requested):

paulb777:
The latest test should address the review comments and pass the tests. I still have some issues with Firebase tests to investigate.

Sources/Core/GTMSessionFetcher.m
```diff
@@ -2667,25 +2671,28 @@ - (void)invokeFetchCallbacksOnCallbackQueueWithData:(nullable NSData *)data
   }
 
   if (handler) {
-    [self invokeOnCallbackQueue:callbackQueue afterUserStopped:NO block:^{
```
paulb777, line 2670: clang-format change only

Sources/Core/GTMSessionFetcher.m
paulb777: I couldn't find a way to get the afterStopped parameter right. I did update the added test to limit to when triggersCompletion is on and there is a userStop occurrence.

thomasvl: Does this still need work or did this get figured out?

thomasvl:
I also did a clang-format on `main`, so if you rebase/merge, the diffs specific to that should go away.

Sources/Core/GTMSessionFetcher.m
paulb777: I've updated to limit the fall through to only when there has been a stop, so yes.

paulb777:
Rebased with no conflicts

Sources/Core/GTMSessionFetcher.m
```diff
@@ -1995,6 +1995,7 @@ - (void)stopFetchReleasingCallbacks:(BOOL)shouldReleaseCallbacks {
   [holdSelf destroyRetryTimer];
```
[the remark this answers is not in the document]
paulb777: Updated

Sources/Core/Public/GTMSessionFetcher/GTMSessionFetcher.h
```diff
@@ -1084,9 +1085,12 @@ __deprecated_msg("implement GTMSessionFetcherAuthorizer instead")
 @property(atomic, readonly, getter=isFetching) BOOL fetching;
 
 // Cancel the fetch of the request that's currently in progress.  The completion handler
-// will not be called.
+// will be called if the property `stopFetchingTriggersCompletionHandler` is `YES`.
```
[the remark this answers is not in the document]
paulb777: Perhaps, it should only be supported on the service, since it doesn't really make sense to change once the fetcher has been created. WDYT?

Sources/Core/GTMSessionFetcher.m
```diff
@@ -2598,9 +2602,11 @@ - (void)invokeOnCallbackUnsynchronizedQueueAfterUserStopped:(BOOL)afterStopped
 - (void)invokeOnCallbackQueue:(dispatch_queue_t)callbackQueue
              afterUserStopped:(BOOL)afterStopped
                         block:(void (^)(void))block {
+
   if (callbackQueue) {
     dispatch_group_async(_callbackGroup, callbackQueue, ^{
-      if (!afterStopped) {
+      if (!afterStopped &&
+          (!self->_stopFetchingTriggersCompletionHandler || !self->_userStoppedFetching)) {
```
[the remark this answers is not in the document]
paulb777: Added sync

Sources/Core/GTMSessionFetcher.m
```diff
@@ -2992,8 +2999,20 @@ - (void)URLSession:(NSURLSession *)session
...
     }
+    userStoppedTriggerCompletion = _userStoppedFetching && _stopFetchingTriggersCompletionHandler;
   }  // @synchronized(self)
 
+  if (userStoppedTriggerCompletion) {
+    NSError *cancelError = [NSError errorWithDomain:kGTMSessionFetcherErrorDomain
+                                               code:GTMSessionFetcherErrorUserCancelled
+                                           userInfo:@{
+                                             NSLocalizedDescriptionKey : @"Operation cancelled",
+                                             NSUnderlyingErrorKey : error.localizedDescription
```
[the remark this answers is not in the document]
paulb777: Fixed

Sources/Core/Public/GTMSessionFetcher/GTMSessionFetcher.h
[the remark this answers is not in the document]
thomasvl: Actually, that might simplify things since it locks in the state so you don't need the synchronization, etc. Would that work for the Firebase usage?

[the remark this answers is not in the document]
paulb777: Actually no - since we want to be able to call a combination of new and old APIs.  However, it does seem reasonable to make `stopFetchingTriggersCompletionHandler` a readonly property and I did that in the latest commit.

Sources/Core/Public/GTMSessionFetcher/GTMSessionFetcherService.h
```diff
@@ -136,9 +136,12 @@ extern NSString *const kGTMSessionFetcherServiceSessionKey;
 // Common method for fetcher creation.
 //
 // -fetcherWithRequest:fetcherClass: may be overridden to customize creation of
-// fetchers.  This is the ONLY method in the GTMSessionFetcher library intended to
+// fetchers.  These are the ONLY method in the GTMSessionFetcher library intended to
 // be overridden.
 - (id)fetcherWithRequest:(NSURLRequest *)request fetcherClass:(Class)fetcherClass;
+- (id)fetcherWithRequest:(NSURLRequest *)request
+              fetcherClass:(Class)fetcherClass
+    stopTriggersCompletion:(BOOL)stopTriggersCompletion;
```
thomasvl, line 144: I was thinking we just make this a property on the service and then you don't need a new api, it just gets copied out like all the other properties that get carried over?

paulb777: That doesn't work if you want to use the service to call both sets of APIs. I suppose I could change Firebase to add a second service just for usage from async APIs?

thomasvl: > Done.

Just checking, did that mean you had another commit to also push?  I don't think one appeared after this comment.

paulb777: Yep. I didn't notice that I typo'd the `git push` command. Thanks.

Sources/Core/Public/GTMSessionFetcher/GTMSessionFetcher.h
```diff
@@ -1084,9 +1085,12 @@ __deprecated_msg("implement GTMSessionFetcherAuthorizer instead")
...
 - (void)stopFetching;
 
+// Call callbacks in `stopFetching`.
```
thomasvl, line 1091: Document that it can't be changed after begin

Sources/Core/GTMSessionFetcher.m
```diff
@@ -3995,6 +4019,21 @@ - (void)setUsingBackgroundSession:(BOOL)flag {
...
+}
+
+- (void)setStopFetchingTriggersCompletionHandler:(BOOL)flag {
+  @synchronized(self) {
+    GTMSessionMonitorSynchronized(self);
+    GTMSESSION_ASSERT_DEBUG(
+        _initialBeginFetchDate == nil,
+        @"stopFetchingTriggersCompletionHandler should not change after fetcher starts");
+
+    _stopFetchingTriggersCompletionHandler = flag;
```
thomasvl, line 4033: This should probably still have a guard so it won't update if _initialBeginFetchDate is non nil.

Sources/Core/GTMSessionFetcher.m
```diff
@@ -3995,6 +4019,21 @@ - (void)setUsingBackgroundSession:(BOOL)flag {
   }  // @synchronized(self)
 }
 
+- (BOOL)stopFetchingTriggersCompletionHandler {
+  return _stopFetchingTriggersCompletionHandler;
+}
+
+- (void)setStopFetchingTriggersCompletionHandler:(BOOL)flag {
+  @synchronized(self) {
```
thomasvl, line 4027: I think you can skip the sync, _initialBeginFetchDate gets set without an sync, so you're probably safe skipping the sync here also. If someone is setting up a fetcher from multiple threads, or starting it while setting it up from multiple threads, they already are in a risky space.

Sources/Core/GTMSessionFetcher.m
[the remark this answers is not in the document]
thomasvl: Sorry, meant line 1981, it currently says:

```
// If shouldReleaseCallbacks is NO then the fetch will be retried so the callbacks
// need to still be retained.
```

Which isn't completely true any more, as the cancel behavior now factors in to when `NO` is used.

Sources/Core/GTMSessionFetcher.m
```diff
@@ -2598,9 +2603,14 @@ - (void)invokeOnCallbackUnsynchronizedQueueAfterUserStopped:(BOOL)afterStopped
 - (void)invokeOnCallbackQueue:(dispatch_queue_t)callbackQueue
              afterUserStopped:(BOOL)afterStopped
                         block:(void (^)(void))block {
+
   if (callbackQueue) {
     dispatch_group_async(_callbackGroup, callbackQueue, ^{
-      if (!afterStopped) {
+      BOOL skipEarlyExit = NO;
+      @synchronized(self) {
```
thomasvl, line 2610: this sync isn't needed now that it can't change after begin.

Sources/Core/GTMSessionFetcher.m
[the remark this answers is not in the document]
paulb777: Updated 1981.

Sources/Core/GTMSessionFetcher.m
paulb777: removed

Sources/Core/GTMSessionFetcher.m
paulb777: done

Sources/Core/Public/GTMSessionFetcher/GTMSessionFetcher.h
paulb777: Done.

paulb777 (commented):
Thanks for the reviews!

Sources/Core/GTMSessionFetcher.m
thomasvl: Something like:

```
  if (_initialBeginFetchDate == nil) {
    _stopFetchingTriggersCompletionHandler = flag;
  } else {
    GTMSESSION_ASSERT_DEBUG(
        0,
        @"stopFetchingTriggersCompletionHandler should not change after fetcher starts");
  }
```

?

paulb777: Updated

Sources/Core/GTMSessionFetcher.m
```diff
@@ -2066,6 +2068,7 @@ - (void)stopFetchReleasingCallbacks:(BOOL)shouldReleaseCallbacks {
         }
       }
     }
+    shouldStopFetcher = !_stopFetchingTriggersCompletionHandler || !_userStoppedFetching;
```
thomasvl, line 2071: Why is this one `!_userStoppedFetcher`?

If `stopFetchingTriggersCompletionHandler` is `NO`, and `stopFetching` is called, then `[service fetcherDidStop:self]` on line 2090 will be skipped, so what actually does that work now?

At first reread, should this be `!_stopFetchingTriggersCompletionHandler && _userStoppedFetching`?  That would make the code continue the current behavior before adding `_stopFetchingTriggersCompletionHandler`.

Sources/Core/Public/GTMSessionFetcher/GTMSessionFetcher.h
```diff
@@ -1084,9 +1085,12 @@ __deprecated_msg("implement GTMSessionFetcherAuthorizer instead")
 @property(atomic, readonly, getter=isFetching) BOOL fetching;
 
 // Cancel the fetch of the request that's currently in progress.  The completion handler
-// will not be called.
+// will be called if the property `stopFetchingTriggersCompletionHandler` is `YES`.
 - (void)stopFetching;
 
+// Call callbacks in `stopFetching`. It cannot be changed once the fetcher starts.
+@property(atomic, assign) BOOL stopFetchingTriggersCompletionHandler;
```
thomasvl, line 1092: Sorry, didn't think of this before – Can you add a block here speaking to the point if the fetcher is going to be use from Swift `async` calls, they this must be set to YES to ensure a cancel doesn't leave the Swift runtime in a bad state?.

Sources/Core/GTMSessionFetcher.m
paulb777: This change hangs the `testFetcherService` and other tests. I'll investigate.

paulb777: After further investigation, I believe the logic was correct. I inverted the condition and it's hopefully easier to understand.

Sources/Core/Public/GTMSessionFetcher/GTMSessionFetcher.h
paulb777: Done

Sources/Core/GTMSessionFetcher.m
thomasvl: Thanks, yea, this does read easier.

thomasvl (approved):

PR #336, opened 2022-11-08
Use local to avoid arc-repeated-use-of-weak warnings.

thomasvl:


thomasvl:
Looks like test flake. 😞

PR #337, opened 2022-11-21
Handle edge case where fetch finishes before cancel completes.

thomasvl:
This is attempting to address what was seen in:
  https://github.com/google/gtm-session-fetcher/actions/runs/3515759994/jobs/5891433576

Given the error, the only option is `error` was `nil`, which likely means the fetch finished. So tweak the code flows, so that doesn't get considered as success and handle where there isn't an underlying error to include in the `userInfo`.


thomasvl:
@paulb777 fyi

PR #341, opened 2023-01-12
Turn down log level of GTMSessionFetcher decorators

bhamiltoncx:
The GTMSessionFetcher decorator logic was logging at DEBUG level on every connection.

This changes it to VERBOSE (which is disabled by default but can be enabled for local debugging).


thomasvl (approved):
Thanks.

Is this the formatting clang-format would do?  (if not, can you run it and re-upload?)

bhamiltoncx:
> Thanks.
> 
> Is this the formatting clang-format would do? (if not, can you run it and re-upload?)

Yes, I formatted the PR with `git clang-format -f`.

Issue #352, opened 2023-05-24
GTMFetcherStandardUserAgentString() is slow


bhamiltoncx:
`GTMFetcherStandardUserAgentString()` blocks the calling thread of `-[GTMSessionFetcherService init]` for several hundred milliseconds when fetching metadata from the bundle and formatting the `User-Agent` string.

The `User-Agent` string isn't needed until the request goes out, so let's:

1) Format this asynchronously off the UI thread if needed
2) Cache it
3) Add a flag so the performance improvement can be validated experimentally

See Google bug b/284149500 for more.

PR #353, opened 2023-05-24
New option to calculate User-Agent string asynchronously and cache it

bhamiltoncx:
This PR introduces a new protocol `GTMUserAgentProvider` which allows calculating the default `User-Agent` header off the UI thread and caching it for re-use.

The new behavior is off by default and requires setting `GTMSessionFetcherService.useStandardUserAgentProvider = YES` (a private `@class` property) so the performance improvement can be experimentally verified.

Tested:
  New tests added. `swift test` passes:
  ```
  Test Suite 'All tests' passed at 2023-05-24 16:31:31.782.
  	 Executed 159 tests, with 0 failures (0 unexpected) in 116.824 (116.834) seconds
  ```

Fixes: #352 
Fixes: b/284149500

### DIFF
--- a/Sources/Core/GTMSessionFetcher.m
+++ b/Sources/Core/GTMSessionFetcher.m
@@ -713,6 +713,25 @@ static GTMSessionFetcherTestBlock _Nullable gGlobalTestBlock;
     }
   }
 
+  if ([fetchRequest valueForHTTPHeaderField:@"User-Agent"] == nil) {
+    id<GTMUserAgentProvider> userAgentProvider = _userAgentProvider;
+    NSString *cachedUserAgent = userAgentProvider.cachedUserAgent;
+    if (cachedUserAgent) {
+      // The User-Agent is already cached in memory, so set it synchronously.
+      [fetchRequest setValue:cachedUserAgent forHTTPHeaderField:@"User-Agent"];
+    } else if (userAgentProvider != nil) {
+      // The User-Agent is not cached in memory. Fetch it asynchronously.
+      [self updateUserAgentAsynchronouslyForRequest:fetchRequest
+                                  userAgentProvider:userAgentProvider
+                                           mayDelay:mayDelay
+                                       mayAuthorize:mayAuthorize
+                                        mayDecorate:mayDecorate];
+      // This method can't continue until the User-Agent header is fetched. The above
+      // method call will re-enter this method later, but with the User-Agent header set.
+      return;
+    }
+  }
+
   NSString *effectiveHTTPMethod = [fetchRequest valueForHTTPHeaderField:@"X-HTTP-Method-Override"];
   if (effectiveHTTPMethod == nil) {
     effectiveHTTPMethod = fetchRequest.HTTPMethod;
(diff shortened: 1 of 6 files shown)


Sources/Core/GTMSessionFetcher.m
```diff
@@ -1004,6 +1022,45 @@ - (NSURLSession *)createSessionWithDelegate:(id<NSURLSessionDelegate>)sessionDel
...
+  __weak __typeof__(self) weakSelf = self;
+  dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+    __strong __typeof__(self) strongSelf = weakSelf;
+    if (!strongSelf) {
+      GTMSESSION_LOG_DEBUG_VERBOSE(@"GTMSessionFetcher deallocated before GTMUserAgentProvider "
+                                   @"fetch dispatched, ignoring.");
+      return;
+    }
+    NSString *userAgent = [userAgentProvider userAgent];
+    GTMSESSION_LOG_DEBUG_VERBOSE(@"Fetched User-Agent string: [%@]", userAgent);
```
thomasvl, line 1043: At this point, would it make sense to assert that the provider now says the value is cached?

Sources/Core/Public/GTMSessionFetcher/GTMSessionFetcher.h
```diff
@@ -511,6 +511,41 @@ typedef void (^GTMSessionFetcherTestResponse)(NSHTTPURLResponse *_Nullable respo
 typedef void (^GTMSessionFetcherTestBlock)(GTMSessionFetcher *fetcherToTest,
                                            GTMSessionFetcherTestResponse testResponse);
 
+// Provides access to a user-agent string calculated on demand.
+@protocol GTMUserAgentProvider <NSObject>
```
thomasvl, line 515: Shouldn't this document that implementor need to make any class conforming to this thread safe?  i.e. - could multiple fetchers ended up trying to call atleast the userAgentCached property while another thread might have already called userAgent?

Sources/Core/GTMSessionFetcherService.m
```diff
@@ -36,6 +47,16 @@ - (BOOL)beginFetchMayDelay:(BOOL)mayDelay
...
 
+// If YES, |userAgentProvider| will default to |GTMStandardUserAgentProvider| for new instances of
+// this class. The User-Agent string will be calculated asynchronously off the calling thread and
+// cached.
+//
+// Otherwise, |userAgentProvider| will default to a |GTMUserAgentStringProvider| containing
+// |GTMFetcherStandardUserAgentString()| fetched on the calling thread (which can block).
+//
+// Defaults to NO.
+@property(class, atomic, assign) BOOL useStandardUserAgentProvider;
```
thomasvl, line 58: Since this isn't public, why provide it?  Doesn't it also mean the only thing ever used is the `NO` code paths?

Sources/Core/GTMSessionFetcherService.m
```diff
@@ -898,6 +924,48 @@ - (void)setSessionDelegateQueue:(NSOperationQueue *)queue {
   }  // @synchronized(self)
 }
 
+- (nullable NSString *)userAgent {
```
thomasvl, line 927: This doesn't seem used?

Sources/Core/Public/GTMSessionFetcher/GTMSessionFetcher.h
bhamiltoncx: Added docs.

Sources/Core/GTMSessionFetcherService.m
```diff
@@ -898,6 +924,48 @@ - (void)setSessionDelegateQueue:(NSOperationQueue *)queue {
...
+  @synchronized(self) {
+    return _userAgentProvider.userAgent;
+  }
+}
+
+- (void)setUserAgent:(nullable NSString *)userAgent {
```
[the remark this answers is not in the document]
bhamiltoncx: Same as above, it's to support the publicly writable property.

Sources/Core/GTMSessionFetcherService.m
```diff
@@ -216,10 +245,7 @@ - (id)fetcherWithRequest:(NSURLRequest *)request fetcherClass:(Class)fetcherClas
   fetcher.skipBackgroundTask = self.skipBackgroundTask;
 #endif
 
-  NSString *userAgent = self.userAgent;
-  if (userAgent.length > 0 && [request valueForHTTPHeaderField:@"User-Agent"] == nil) {
-    [fetcher setRequestValue:userAgent forHTTPHeaderField:@"User-Agent"];
-  }
+  fetcher.userAgentProvider = self.userAgentProvider;
```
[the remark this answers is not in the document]
bhamiltoncx: I thought about it, but I don't think it's worth optimizing that, since as you said fetchers are typically started right after they're created.

Sources/Core/GTMSessionFetcher.m
bhamiltoncx: Done.

thomasvl (approved):

Issue #367, opened 2024-01-03
-[GTMSessionFetcherServiceTest testMultipleDecoratorsSynchronous] fails when run alone


bhamiltoncx:
```
% swift test --filter GTMSessionFetcherServiceTest/testMultipleDecoratorsSynchronous
Building for debugging...
Build complete! (0.12s)
Test Suite 'Selected tests' started at 2024-01-03 15:02:08.423.
Test Suite 'GTMSessionFetcherPackageTests.xctest' started at 2024-01-03 15:02:08.424.
Test Suite 'GTMSessionFetcherServiceTest' started at 2024-01-03 15:02:08.424.
Test Case '-[GTMSessionFetcherServiceTest testMultipleDecoratorsSynchronous]' started.
2024-01-03 15:02:08.427 xctest[4295:1046054] GTMSessionFetcher 0x130f16fd0 (https://www.html5zombo.com) was already running
/Users/benhamilton/Developer/OpenSource/gtm-session-fetcher/UnitTests/GTMSessionFetcherServiceTest.m:1333: error: -[GTMSessionFetcherServiceTest testMultipleDecoratorsSynchronous] : ((FetcherHeadersWithoutUserAgent(fetcher)) equal to (expectedHeaders)) failed: ("(null)") is not equal to ("{
    baz = blech;
    foo = bar;
    quux = corge;
}")
Test Case '-[GTMSessionFetcherServiceTest testMultipleDecoratorsSynchronous]' failed (0.053 seconds).
Test Case '-[GTMSessionFetcherServiceTest testMultipleDecoratorsSynchronousFetchedDataAndError]' started.
Test Case '-[GTMSessionFetcherServiceTest testMultipleDecoratorsSynchronousFetchedDataAndError]' passed (0.001 seconds).
Test Case '-[GTMSessionFetcherServiceTest testMultipleDecoratorsSynchronousWithErrorShouldNotCallSubsequentDecorators]' started.
Test Case '-[GTMSessionFetcherServiceTest testMultipleDecoratorsSynchronousWithErrorShouldNotCallSubsequentDecorators]' passed (0.001 seconds).
Test Suite 'GTMSessionFetcherServiceTest' failed at 2024-01-03 15:02:08.479.
	 Executed 3 tests, with 1 failure (0 unexpected) in 0.054 (0.055) seconds
Test Suite 'GTMSessionFetcherPackageTests.xctest' failed at 2024-01-03 15:02:08.479.
[...]

Issue #368, opened 2024-01-03
-[GTMSessionFetcherFetchingTest testFetcherRedirectURLHandling] fails on macOS 14.2.1


bhamiltoncx:
Looks like macOS Sonoma (14.2) is more strict about invalid URL schemes like `invalid://`:

```
% swift test --filter GTMSessionFetcherFetchingTest/testFetcherRedirectURLHandling
Building for debugging...
Build complete! (0.13s)
Test Suite 'Selected tests' started at 2024-01-03 15:03:29.899.
Test Suite 'GTMSessionFetcherPackageTests.xctest' started at 2024-01-03 15:03:29.900.
Test Suite 'GTMSessionFetcherFetchingTest' started at 2024-01-03 15:03:29.900.
Test Case '-[GTMSessionFetcherFetchingTest testFetcherRedirectURLHandling]' started.
/Users/benhamilton/Developer/OpenSource/gtm-session-fetcher/UnitTests/GTMSessionFetcherFetchingTest.m:2786: error: -[GTMSessionFetcherFetchingTest testFetcherRedirectURLHandling] : ((redirectURL) equal to ([NSURL URLWithString:testCase[2]])) failed: ("invalid://redirect_host/") is not equal to ("https://redirect_host/")
/Users/benhamilton/Developer/OpenSource/gtm-session-fetcher/UnitTests/GTMSessionFetcherFetchingTest.m:2786: error: -[GTMSessionFetcherFetchingTest testFetcherRedirectURLHandling] : ((redirectURL) equal to ([NSURL URLWithString:testCase[2]])) failed: ("http:%5Curl") is not equal to ("http://original_host/")
Test Case '-[GTMSessionFetcherFetchingTest testFetcherRedirectURLHandling]' failed (0.034 seconds).
Test Suite 'GTMSessionFetcherFetchingTest' failed at 2024-01-03 15:03:29.934.
	 Executed 1 test, with 2 failures (0 unexpected) in 0.034 (0.035) seconds
Test Suite 'GTMSessionFetcherPackageTests.xctest' failed at 2024-01-03 15:03:29.934.
	 Executed 1 test, with 2 failures (0 unexpected) in 0.034 (0.035) seconds
Test Suite 'Selected tests' failed at 2024-01-03 15:03:29.934.
	 Executed 1 test, with 2 failures (0 unexpected) in 0.034 (0.036) seconds
```

PR #369, opened 2024-01-03
[UnitTests] Fix order-dependent synchronous decorator tests

bhamiltoncx:
The decorator tests which used synchronous decorators started failing if they were run alone (before any of the other tests).

This broke in
https://github.com/google/gtm-session-fetcher/commit/d469daaafc456f2fc4625918987c0345761a2648 , when I enabled `GTMStandardUserAgentProvider` by default in fetchers which don't otherwise specify a user-agent provider.

The issue is that `GTMStandardUserAgentProvider` itself is asynchronous, but only the first time it's ever invoked (after that, it caches its result in a process-wide cache).

The synchronous decorator tests assumed they'd always stay synchronous, but this wasn't the case if they used `GTMStandardUserAgentProvider` and were the first tests to run in the process.

This fixes the issue by always providing a `GTMUserAgentStringProvider` for the synchronous decorator tests.

Tested:
  `swift test` passes (except known failure in #368, which is already failing in upstream on macOS 14.2.1).
  Also ran test
  `-[GTMSessionFetcherServiceTest testMultipleDecoratorsSynchronous]`
  by itself and ensured it failed before this change and passed after
  this change.

Fixes: #367

### DIFF
--- a/UnitTests/GTMSessionFetcherServiceTest.m
+++ b/UnitTests/GTMSessionFetcherServiceTest.m
@@ -1185,6 +1185,10 @@ static bool IsCurrentProcessBeingDebugged(void) {
       [GTMSessionFetcherService mockFetcherServiceWithFakedData:nil fakedError:nil];
   [service addDecorator:decorator];
   GTMSessionFetcher *fetcher = [service fetcherWithURLString:@"https://www.html5zombo.com"];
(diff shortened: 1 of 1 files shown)


thomasvl (approved):

thomasvl:
Thanks for the fix!

PR #397, opened 2024-06-17
[Swift] Annotate GTMFetcherDecoratorFetcherWillStartCompletionHandler with _Nullable_result

bhamiltoncx:


Sources/Core/Public/GTMSessionFetcher/GTMSessionFetcher.h
```diff
@@ -605,7 +605,7 @@ NSData *_Nullable GTMDataFromInputStream(NSInputStream *inputStream, NSError **o
 #endif
 
 // Completion handler passed to -[GTMFetcherDecoratorProtocol fetcherWillStart:completionHandler:].
-typedef void (^GTMFetcherDecoratorFetcherWillStartCompletionHandler)(NSURLRequest *_Nullable,
+typedef void (^GTMFetcherDecoratorFetcherWillStartCompletionHandler)(NSURLRequest *_Nullable_result,
```
thomasvl, line 608: Make this a TODO instead since it is a breaking change, and next time we have to do a major version bump we can do it.

bhamiltoncx: Done, filed https://github.com/google/gtm-session-fetcher/issues/398 to track the major version bump and added a TODO.

thomasvl (approved):

Issue #402, opened 2024-09-20
Calling +[GTMSessionUploadFetcher fetcherWithSessionIdentifier:] incorrectly returns GTMSessionFetcher


bhamiltoncx:
[`+[GTMSessionUploadFetcher fetcherWithSessionIdentifier:]`](https://github.com/google/gtm-session-fetcher/blob/dc9988c86d894246af15af21f8ed4a01acf99a50/Sources/Core/Public/GTMSessionFetcher/GTMSessionFetcher.h#L790) is declared to return `instancetype`, so it should return `GTMSessionUploadFetcher*`.

In reality, [that method has no implementation in `GTMSessionUploadFetcher.m`, so it always returns `GTMSessionFetcher*`](https://github.com/google/gtm-session-fetcher/blob/dc9988c86d894246af15af21f8ed4a01acf99a50/Sources/Core/GTMSessionFetcher.m#L319). That means calling `GTMSessionUploadFetcher`-only methods on the resulting object will crash, since they're not implemented.

The method [`+[GTMSessionUploadFetcher uploadFetcherWithSessionIdentifier:]`](https://github.com/google/gtm-session-fetcher/blob/dc9988c86d894246af15af21f8ed4a01acf99a50/Sources/Core/Public/GTMSessionFetcher/GTMSessionUploadFetcher.h#L124) [is implemented to return the correct type](https://github.com/google/gtm-session-fetcher/blob/dc9988c86d894246af15af21f8ed4a01acf99a50/Sources/Core/GTMSessionUploadFetcher.m#L335), so we can implement `+[GTMSessionUploadFetcher fetcherWithSessionIdentifier:]` to thunk through to that method.

PR #403, opened 2024-09-20
Implement +[GTMSessionUploadFetcher fetcherWithSessionIdentifier:]

bhamiltoncx:
`+[GTMSessionUploadFetcher fetcherWithSessionIdentifier:]` is declared to return `instancetype`, so it should return `GTMSessionUploadFetcher*`.

In reality, that method has no implementation in `GTMSessionUploadFetcher.m`, so it always returns `GTMSessionFetcher*`. That means calling `GTMSessionUploadFetcher`-only methods on the resulting object will crash, since they're not implemented.

The method `+[GTMSessionUploadFetcher uploadFetcherForSessionIdentifier:]` is implemented to return the correct type, so this PR implements `+[GTMSessionUploadFetcher fetcherWithSessionIdentifier:]` to thunk through to that method.

Fixes: #402

### DIFF
--- a/Sources/Core/GTMSessionUploadFetcher.m
+++ b/Sources/Core/GTMSessionUploadFetcher.m
@@ -332,6 +332,10 @@ NSString *const kGTMSessionFetcherUploadInitialBackoffStartedNotification =
   return gUploadFetcherPointerArrayForBackgroundSessions;
 }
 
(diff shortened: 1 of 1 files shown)


thomasvl:
We can ignore the visionOS failures.  Github dropped the SDK from their runners and `main` has since been updated to stop trying to run the tests as a result.

thomasvl (approved):

PR #411, opened 2024-10-11
Bring back visionOS for SwiftPM testing.

thomasvl:


thomasvl:
The CocoaPods testing seems to still require the simulators, which aren't installed on the image we're using.

PR #426, opened 2025-01-02
Protect the logging state storage for threaded access.

thomasvl:
The thread sanitizer properly flagged a case of this, so generally fix the problem. Since multiple fetches could happen in parallel, some of these can be lazy initialized in parallel, and thus exposed to threading issues, protect the access to avoid any issues.

### DIFF
--- a/Sources/Core/GTMSessionFetcherLogging.m
+++ b/Sources/Core/GTMSessionFetcherLogging.m
@@ -66,7 +66,9 @@
 
 @implementation GTMSessionFetcher (GTMSessionFetcherLogging)
 
-// fetchers come and fetchers go, but statics are forever
+// This should only be accessed within @synchronized([GTMSessionFetcher class]) blocks
+// to ensure thread safety. Even if not manually toggled, multiple fetches can lazy
+// initialize things in parallel causing issues.
 static BOOL gIsLoggingEnabled = NO;
 static BOOL gIsLoggingToFile = YES;
 static NSString *gLoggingDirectoryPath = nil;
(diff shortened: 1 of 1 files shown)


Sources/Core/GTMSessionFetcherLogging.m
```diff
@@ -66,7 +66,9 @@ - (void)inputStream:(GTMReadMonitorInputStream *)stream
 
 @implementation GTMSessionFetcher (GTMSessionFetcherLogging)
 
-// fetchers come and fetchers go, but statics are forever
+// This should only be accesses within @synchronized([GTMSessionFetcher class]) blocks
```
dmaclach, line 69: s/accesses/accessed/

Sources/Core/GTMSessionFetcherLogging.m
```diff
@@ -75,141 +77,165 @@ @implementation GTMSessionFetcher (GTMSessionFetcherLogging)
...
-      NSString *const kGTMLogFolderName = @"GTMHTTPDebugLogs";
-      NSString *logsFolderPath = [desktopPath stringByAppendingPathComponent:kGTMLogFolderName];
+      NSString *desktopPath = paths.firstObject;
+      if (desktopPath) {
+        NSString *const kGTMLogFolderName = @"GTMHTTPDebugLogs";
+        NSString *logsFolderPath = [desktopPath stringByAppendingPathComponent:kGTMLogFolderName];
+
+        NSFileManager *fileMgr = [NSFileManager defaultManager];
+        BOOL isDir;
+        BOOL doesFolderExist = [fileMgr fileExistsAtPath:logsFolderPath isDirectory:&isDir];
```
dmaclach, line 109: Instead of checking and then creating, why don't you just create it every time? Get rid of a minor race condition.

Sources/Core/GTMSessionFetcherLogging.m
```diff
@@ -75,141 +77,165 @@ @implementation GTMSessionFetcher (GTMSessionFetcherLogging)
...
+        if (!doesFolderExist) {
+          // make the directory
+          doesFolderExist = [fileMgr createDirectoryAtPath:logsFolderPath
+                               withIntermediateDirectories:YES
+                                                attributes:nil
+                                                     error:NULL];
+          if (doesFolderExist) {
```
dmaclach, line 116: What are we doing in the unlikely case that this fails?

Sources/Core/GTMSessionFetcherLogging.m
```diff
@@ -75,141 +77,165 @@ @implementation GTMSessionFetcher (GTMSessionFetcherLogging)
...
-      NSString *origLogDir = logDirectory;
-      for (int ctr = 2; ctr < 20; ++ctr) {
-        if (![fileMgr fileExistsAtPath:logDirectory]) break;
-
-        // append a digit
-        logDirectory = [origLogDir stringByAppendingFormat:@"_%d", ctr];
+    if (gIsLoggingToFile) {
+      NSFileManager *fileMgr = [NSFileManager defaultManager];
+      // Be sure that the first time this app runs, it's not writing to a preexisting folder
+      static BOOL gShouldReuseFolder = NO;
```
dmaclach, line 154: Do you have another thread sanitizer issue here with an unprotected global?

thomasvl: How?  This whole body is within a sync block and only use here.

Sources/Core/GTMSessionFetcherLogging.m
thomasvl: This is the existing flow, so I'd sorta like to not make other changes at the same time we fold in locks.  We can always revisit it in the future, but minimizing scope of changes per commit seems worth while.  Thoughts?

Sources/Core/GTMSessionFetcherLogging.m
thomasvl: Same.

Sources/Core/GTMSessionFetcherLogging.m
thomasvl: Done.

dmaclach (approved):
LGTM

PR #447, opened 2025-03-31
Add basic Bazel build support for gtm-session-fetcher

### LINKED ISSUE #446: [Feature Request] Add Bazel Support
Currently, gtm-session-fetcher supports CocoaPods and Swift Package Manager (SPM). It would be beneficial to add support for Bazel as well.

bc-lee:
Fixes #446


MODULE.bazel
```diff
@@ -0,0 +1,12 @@
+module(name = "gtm_session_fetcher", version = "4.4.0")
```
thomasvl, line 1: Forgot to ask, does this version need to be raised each time we make a release?

MODULE.bazel.lock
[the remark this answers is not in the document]
bc-lee: Right, I forgot to add it to the .gitignore. It’s fixed.

MODULE.bazel
bc-lee: Right. We need to update the `MODULE.bazel` version whenever a new release is created.
This version value is also important for uploading to BCR (https://registry.bazel.build/).

.bazelversion
```diff
@@ -0,0 +1 @@
+8.2.0
```
thomasvl, line 1: Is it standard practice to pin this?  Won't bazelisk otherwise default to the current "released" version?  Would it be better to just stay floating so it tracks what is current and would be one less thing to update?

MODULE.bazel
thomasvl: Thanks, we'll want to eventually roll that into `update_version.py` then so it doesn't get missed.

.bazelversion
bc-lee: Yes, it's generally considered standard practice for Bazel-based projects to pin the Bazel version using a `.bazelversion` file. This helps ensure deterministic and reproducible builds. Bazel projects often depend on multiple rulesets, and using different Bazel versions can lead to incompatibilities or unexpected build failures.

While we could configure CI to test against multiple Bazel versions using `USE_BAZEL_VERSION` with Bazelisk—potentially catching breakages with the latest release—for this initial step, I believe it's reasonable to pin a known-good version and align CI with it. This helps ensure consistency between local development and CI environments and avoids surprises from upstream Bazel changes.

thomasvl (approved):
Thanks again for the contribution!

PR #448, opened 2025-04-01
Expand the tests for a stopping a single fetcher.

thomasvl:
- Chain the non service test to call the the others to reduce chance for coding errors.
- Add support for calling `stopFetching` even before beginning the fetch.
- Add support to `TestAuthorizer` to have a block called on a work queue before the authorization completes.
- Ensure the authorizer isn't called if the stop is done before the fetch starts.
- Ensure the authorizer is called if the stop happens after starting the fetch.
- When delaying the authorizer, control exactly when it triggers to control where in the flow of starting a fetcher things things happen.

This also remove some of the "sleep" delays in the tests, to speed up these tests.

### DIFF
--- a/UnitTests/GTMSessionFetcherFetchingTest.m
+++ b/UnitTests/GTMSessionFetcherFetchingTest.m
@@ -1682,91 +1682,143 @@ NSString *const kGTMGettysburgFileName = @"gettysburgaddress.txt";
 #endif
 }
 
+typedef NS_ENUM(NSInteger, TestAuthorizerMode) {
+  TestAuthorizerModeNone = 0,
+  TestAuthorizerModeSync,
+  TestAuthorizerModeAsync,
+  TestAuthorizerModeWaitPreSleep,
+  TestAuthorizerModeWaitPreStop,
+  TestAuthorizerModeWaitPostStop,
+};
+
 - (void)testDelayedCancelFetchWithCallback {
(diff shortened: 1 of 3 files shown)


UnitTests/GTMSessionFetcherFetchingTest.m
```diff
@@ -1783,28 +1835,81 @@ - (void)internalCancelFetchWithCallback:(unsigned int)sleepTime
...
+        break;
+      case TestAuthorizerModeSync:
+        fetcher.authorizer = [TestAuthorizer syncAuthorizer];
+        break;
+      case TestAuthorizerModeAsync:
+        fetcher.authorizer = [TestAuthorizer asyncAuthorizer];
+        break;
+      case TestAuthorizerModeWaitPreSleep:
+        XCTAssertGreaterThan(sleepTime, 0,
+                             @"Pre sleep when stop was before begin doensn't make sense.");
```
dmaclach, line 1860: "PreSleep when 'stop' was before 'begin' doesn't make sense."

Makes it easier to read if you put the quote in there

dmaclach (dismissed):

UnitTests/GTMSessionFetcherFetchingTest.m
thomasvl: Done

PR #455, opened 2025-04-07
Add testing for UserAgent Provider and `-stopFetching`

thomasvl:
The is broken into a few chucks to hopefully be easier to review.

- Refactor some existing helpers into a reusable location, updating the existing usages.
- Move the blocking/unblock model into the helper and updated existing usages to use it.
- Fix an existing test that was wrong, the call shouldn't have been happening, confirm it wasn't happening.
- The add all the new tests. NOTE: one of these new tests is failing. That will be address in a follow up, I didn't want to make this any larger by working on that in the same commit chain.

### DIFF
--- a/UnitTests/GTMSessionFetcherFetchingTest.m
+++ b/UnitTests/GTMSessionFetcherFetchingTest.m
@@ -2588,6 +2588,175 @@ typedef void (^StopFetchingCallbackTestBlock)(GTMSessionFetcher *fetcher);
   [self testStopFetchWithCallback_ImmediateStop_AsyncAuthPostStop];
 }
 
+- (void)testStopFetchWithCallback_DelayedStop_UAProviderPreSleep {
+  // Using a UAProvider, trigger the provider after starting the fetch, then wait 1s before calling
+  // `-stopFetching`.
+
+  XCTSkip(@"not currently passing");
+  XCTestExpectation *providerExpect =
+      [self expectationWithDescription:@"Expect for UAProvider block"];
+
+  [self runStopFetchingCallbackTestWithNotifications:NO
+      preBegin:^(GTMSessionFetcher *fetcher) {
+        fetcher.userAgentProvider =
+            [[TestUserAgentBlockProvider alloc] initWithBlockedTimeout:1
+                                                        userAgentBlock:^{
+                                                          [providerExpect fulfill];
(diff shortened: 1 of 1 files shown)


UnitTests/GTMSessionFetcherFetchingTest.m
```diff
@@ -3419,7 +3423,30 @@ - (instancetype)initWithUserAgentBlock:(UserAgentBlock)userAgentBlock {
...
+  NSAssert(_waitSeconds, @"This was not a blocked UAProvider");
+  dispatch_semaphore_signal(_semaphore);
+  _waitSeconds = 0;
+}
+
 - (NSString *)userAgent {
+  if (_waitSeconds) {
+    intptr_t waitResult = dispatch_semaphore_wait(
+        _semaphore, dispatch_time(DISPATCH_TIME_NOW, _waitSeconds * NSEC_PER_SEC));
+    XCTAssertEqual(0, waitResult);
```
dmaclach, line 3617: May be useful for debugging to add a `,@"Timed out after %d seconds", __waitSeconds);`

UnitTests/GTMSessionFetcherFetchingTest.h
```diff
@@ -119,6 +120,20 @@ extern NSString *const kGTMGettysburgFileName;
...
 @end
 
+#pragma mark - User Agent Caching testing helper
+
+typedef NSString *_Nonnull (^UserAgentBlock)(void);
+
+@interface TestUserAgentBlockProvider : NSObject <GTMUserAgentProvider>
+
+@property(atomic, copy) NSString *cachedUserAgent;
+@property(atomic) UserAgentBlock userAgentBlock;
```
dmaclach, line 130: should probably mark the @property as copy (I know that it is by default for a block, but it makes it clear)

dmaclach (dismissed):

UnitTests/GTMSessionFetcherFetchingTest.m
thomasvl: Done, and also did that to the TestAuthorizer support.

UnitTests/GTMSessionFetcherFetchingTest.h
thomasvl: done.

PR #456, opened 2025-04-08
Ensure notifications are posted in order

thomasvl:


Sources/Core/GTMSessionFetcher.m
```diff
@@ -2954,22 +2955,51 @@ - (void)invokeFetchCallbacksOnCallbackQueueWithData:(nullable NSData *)data
...
-    postBlock();
-  } else {
-    dispatch_async(dispatch_get_main_queue(), postBlock);
+    if (_pendingNotifications == 0 && [NSThread isMainThread]) {
+      canPostNow = YES;
+    } else {
+      // This one will get deferred, so increase the counter for deferred notifications.
+      _pendingNotifications++;
+    }
+    if (!canPostNow) {
```
dmaclach, line 2980: does all of this below need to be in the @synchronized block? If looks like if you took it out you could do an if (canPostNow) {...} else {...}

dmaclach (approved):

Sources/Core/GTMSessionFetcher.m
thomasvl: You are right, it doesn't have to be.

PR #462, opened 2025-04-09
Add a check for `-stopFetching` before the final startup.

thomasvl:
When `-stopFetching` is called really early in the startup process and there is no configuration that might cause a delay (authorizer, UAProvider, etc.), then this ensure the fetch never really begins and if requested the completion is triggered.

### DIFF
--- a/Sources/Core/GTMSessionFetcher.m
+++ b/Sources/Core/GTMSessionFetcher.m
@@ -872,6 +872,26 @@ static GTMSessionFetcherTestBlock _Nullable gGlobalTestBlock;
     }
   }
 
+  BOOL stopped;
+  @synchronized(self) {
+    GTMSessionMonitorSynchronized(self);
+    stopped = _userStoppedFetching;
+  }
+  if (stopped) {
+    GTMSESSION_ASSERT_DEBUG(_delayState == kDelayStateNotDelayed,
+                            @"Unexpected internal state: %lu", (unsigned long)_delayState);
+    // We end up here if someone calls `stopFetching` from another thread/queue while
+    // the fetch was being started up, so while `stopFetching` did the needed shutdown
+    // we have to ensure the requested callback was triggered.
+    if (self.stopFetchingTriggersCompletionHandler) {
+      NSError *error = [NSError errorWithDomain:kGTMSessionFetcherErrorDomain
+                                           code:GTMSessionFetcherErrorUserCancelled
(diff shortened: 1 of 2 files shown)


Sources/Core/GTMSessionFetcher.m
```diff
@@ -872,6 +872,26 @@ - (void)beginFetchMayDelay:(BOOL)mayDelay
     }
   }
 
+  BOOL stopped;
```
thomasvl, line 875: This code is basically a subset of `-startDelayState:`, it would work to reuse that with the state being passed being `kDelayStateNotDelayed`, but I wasn't sure if that would be a little too cryptic. We could comment it to make it clear the call is being done for code reuse alone.  Thoughts?

dmaclach: I think it's better to leave it out here explicitly. This code is complicated enough that keeping it here makes it easier for me to follow than nesting it in.

Sources/Core/GTMSessionFetcher.m
```diff
@@ -872,6 +872,26 @@ - (void)beginFetchMayDelay:(BOOL)mayDelay
...
+  if (stopped) {
+    GTMSESSION_ASSERT_DEBUG(_delayState == kDelayStateNotDelayed,
+                            @"Unexpected internal state: %lu", (unsigned long)_delayState);
+    // We end up here if someone calls `stopFetching` from another thread/queue while
+    // the fetch was being started up, so while `stopFetching` did the needed shutdown
+    // we have to ensure the requested callback was triggered.
+    if (self.stopFetchingTriggersCompletionHandler) {
+      NSError *error = [NSError errorWithDomain:kGTMSessionFetcherErrorDomain
+                                           code:GTMSessionFetcherErrorUserCancelled
+                                       userInfo:nil];
```
dmaclach, line 889: Do we ever bother with supplying any maybe useful userInfo like the URL being fetched (or cancelled in this case).

dmaclach (approved):

Sources/Core/GTMSessionFetcher.m
thomasvl: Not currently, but I have a follow up item in my notes to maybe revisit this in the future.

(In most contexts the `fetcher` is also right there at the `begin...` call, so they could use it in their handler also)

PR #481, opened 2025-04-28
Avoid races in begin fetch

thomasvl:
This is split into two commits, looking at them individually will hopefully be easiest. The first closes the main issue, the second does some renaming/cleanup after the fact due to how this all has evolved while nailing things down.

### DIFF
--- a/Sources/Core/GTMSessionFetcher.m
+++ b/Sources/Core/GTMSessionFetcher.m
@@ -125,15 +125,17 @@ NS_ASSUME_NONNULL_END
 #endif
 
 // Internal tracking of the delayed state.
-//
-// This is currently not complete, it is being worked in to better track things
-// to ensure `stopFetching` always triggers callback when requested.
 typedef NS_ENUM(NSUInteger, GTMSessionFetcherDelayState) {
   kDelayStateNotDelayed = 0,
+
+  // Parts of startup that can result in the fetcher completing startup at some later time.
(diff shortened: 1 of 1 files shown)


Sources/Core/GTMSessionFetcher.m
```diff
@@ -498,22 +499,22 @@ - (void)beginFetchWithDelegate:(nullable id)target
...
+- (BOOL)startingState:(GTMSessionFetcherStartingState)newState __attribute__((objc_direct)) {
   BOOL stopped = NO;
   @synchronized(self) {
     GTMSessionMonitorSynchronized(self);
 
     GTMSESSION_ASSERT_DEBUG(
-        (_delayState == kDelayStateNotDelayed || _delayState == kDelayStateStartingUp),
-        @"Unexpected internal state: %lu", (unsigned long)_delayState);
+        (_startingState == kStartingStateNone || _startingState == kStartingStateStartingUp),
+        @"Unexpected internal state: %lu", (unsigned long)_startingState);
```
dmaclach, line 512: nit: "Unexpected starting state"? Just to help give folks an idea what to look for on failure.

dmaclach (dismissed):

Sources/Core/GTMSessionFetcher.m
thomasvl: Done.

PR #500, opened 2025-07-07
Basically revert #409

thomasvl:
This is a breaking change in that it changes the minOS.

Before moving to fully match Firebase, start by basically rolling back what was needed in #409 when we had to keep iOS 12 support.


thomasvl:
@paulb777 fyi.

PR #503, opened 2025-07-11
Fix potential data races with statis in the logging code

thomasvl:
- Move `gResponseCounter` over to `atomic_int` so it is a thread safe count. This is what got flagged originally by someone.
- Comment on all the end `@synchronized` scopes to make code scanning easier.
- Drop `gReportedLoggingPath` and just use `dispatch_once` to make that threadsafe.
- `+processNameLogPrefix` is called from one `@synchronized` block but is also a documented api and called from a place without the protection, so add `@synchronized` to its implementation to protect `gPrefix`.

### DIFF
--- a/Sources/Core/GTMSessionFetcherLogging.m
+++ b/Sources/Core/GTMSessionFetcherLogging.m
@@ -17,6 +17,7 @@
 #error "This file requires ARC support."
 #endif
 
+#include <stdatomic.h>
 #include <unistd.h>
 
 #import "GTMSessionFetcher/GTMSessionFetcherLogging.h"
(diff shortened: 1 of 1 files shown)


Sources/Core/GTMSessionFetcherLogging.m
```diff
@@ -235,16 +236,18 @@ + (NSString *)loggingDateStamp {
...
 }
 
 + (NSString *)processNameLogPrefix {
-  static NSString *gPrefix = nil;
-  if (!gPrefix) {
-    NSString *processName = [self loggingProcessName];
-    gPrefix = [[NSString alloc] initWithFormat:@"%@_log_", processName];
-  }
-  return gPrefix;
+  @synchronized([GTMSessionFetcher class]) {
```
dmaclach, line 243: This looks like a great case for dispatch_once instead?

Sources/Core/GTMSessionFetcherLogging.m
```diff
@@ -385,8 +388,8 @@ - (void)logFetchWithError:(NSError *)error {
   //
   // we'll use a local variable since this routine may be reentered while waiting for formatting
   // to be completed.
-  static int gResponseCounter = 0;
-  int responseCounter = ++gResponseCounter;
+  static atomic_int gResponseCounter = 0;
+  int responseCounter = atomic_fetch_add(&gResponseCounter, 1);
```
dmaclach, line 392: ++gResponseCounter should be sufficient here https://en.cppreference.com/w/c/language/atomic.html

thomasvl: Done

PR #578, opened 2026-09-14
Fix invalid cast in GTMSessionFetcherService for proxied delegates (#190)

### LINKED ISSUE #190: Invalid cast in GTMSessionFetcherService
Hello

This issue was described in several tickets but solution is still not there.

There are several third-party libraries that swizzle `URLSession`'s `delegate` with their own proxy objects. For example TrustKit and SplunkMint do that. However `GTMSessionFetcherService` completely ignores this fact and hard-casts these proxy objects to its own class - `GTMSessionFetcherSessionDelegateDispatcher`: 

https://github.com/google/gtm-session-fetcher/blob/c879a387e0ca4abcdff9e37eb0e826f7142342b1/Source/GTMSessionFetcherService.m#L396

```
    BOOL hasDispatcher = (fetcherDelegate != nil &&
                          ![fetcherDelegate isKindOfClass:[GTMSessionFetcher class]]);
    if (hasDispatcher) {
      GTMSESSION_ASSERT_DEBUG([fetcherDelegate isKindOfClass:[GTMSessionFetcherSessionDelegateDispatcher class]],
                              @"Fetcher delegate class: %@", [fetcherDelegate class]);
      return (GTMSessionFetcherSessionDelegateDispatcher *)fetcherDelegate;
    }
```

Obviously later we get crashes in different places due to "unrecognized selector sent to instance" exceptions.

The "dumb" solution is to check the class of object before casting it:

```
      if ([fetcherDelegate isKindOfClass:[GTMSessionFetcherSessionDelegateDispatcher class]]) {
        return (GTMSessionFetcherSessionDelegateDispatcher *)fetcherDelegate;
      }
      else {
        return nil;
      }
```

But having almost zero knowledge of how GTMSessionFetcher works, it's hard for me to understand if this code breaks something else. Any advice from the team?

Thanks,
Vlad

rksharma-owg:
### Fix invalid cast in `GTMSessionFetcherService` ([#190](https://github.com/google/gtm-session-fetcher/issues/190))

- **Issue**: In `-[GTMSessionFetcherService delegateDispatcherForFetcher:]`, when the session delegate is proxied or wrapped by third-party SDKs (such as TrustKit, Datadog, New Relic), `![fetcherDelegate isKindOfClass:[GTMSessionFetcher class]]` evaluates to `YES`. The code previously assumed that any non-fetcher delegate was a `GTMSessionFetcherSessionDelegateDispatcher` and forced a cast, causing crashes in production (`unrecognized selector sent to instance`).
- **Fix**: Check `[fetcherDelegate isKindOfClass:[GTMSessionFetcherSessionDelegateDispatcher class]]` directly and safely return `nil` if it is not a dispatcher.
- **Unit Test**: Added a test case in `UnitTests/GTMSessionFetcherServiceTest.m` verifying that arbitrary delegates return `nil` without crashing or triggering assertions.

Fixes #190


thomasvl:
Do these really need to be tied into a single PR?  It's far safer to address things in individual commits incase something comes up later and need to be revisited/rolled back/etc.

rksharma-owg:
That makes complete sense! Splitting them into individual PRs makes review, testing, and rollbacks much safer and cleaner.

I have updated this PR (#578) to focus exclusively on Issue #190 (fixing the invalid delegate cast in `GTMSessionFetcherService`). I will submit the other fixes as separate, individual PRs.

thomasvl:
fyi - I'm not sure the test here is worth it.

As to the core change itself, for all the reasons discussed on the original issue, this still seems risky. When folks swizzle and don't forward things, they are fundamentally breaking things the library is built on. So removal of the assert lets those things slide through that much easier.

rksharma-owg:
The current patch removes the debug invariant and lets an incompatible delegate proxy continue without ensuring the forwarding contract works. The added test substitutes an arbitrary delegate, so it does not establish the third-party proxy behavior discussed in #190. I’m closing this rather than weakening the invariant without an endorsed, representative fix.

PR #583, opened 2026-09-14
Support testBlock on restarted upload fetchers without querying network (#185)

### LINKED ISSUE #185: Upload fetchers with a testBlock set can still query the network.
When a testBlock is provided to GTMSessionUploadFetcher but the [upload is considered to be restarted](https://github.com/google/gtm-session-fetcher/blob/96b3d35b30a983f152d4cc11db0adb45adf363eb/Source/GTMSessionUploadFetcher.m#L556), the upload fetch will [query the initial offset](https://github.com/google/gtm-session-fetcher/blob/96b3d35b30a983f152d4cc11db0adb45adf363eb/Source/GTMSessionUploadFetcher.m#L896) from the network even if the test block will supply test data.

Should provide means to override this query.

rksharma-owg:
### Support `testBlock` on restarted upload fetchers ([#185](https://github.com/google/gtm-session-fetcher/issues/185))

**Problem**

A fetcher restarted with `uploadFetcherWithLocation:` always queried the upload server for its initial offset. That bypassed a supplied `testBlock`, so restarted-upload tests could make a real network request. Child chunk fetchers also did not inherit the test block.

**Root cause**

Restarted fetchers do not have an initial request. The upload fetcher returned through the offset-query path before the superclass could invoke `testBlock`.

**Solution**

- Determine `hasTestBlock` once, with `GTM_DISABLE_FETCHER_TEST_BLOCK` handled at compile time.
- Keep the existing offset query for restarted uploads without a test block.
- Only for the test-block path, synthesize the request that the superclass needs to invoke `testBlock`.
- Propagate the test block to child chunk fetchers when test blocks are enabled.
- Add a restarted-upload regression test using an HTTPS fixture URL.

**Testing**

Local:

- `swift build --configuration debug`
- `swift build --configuration debug -Xcc -DGTM_DISABLE_FETCHER_TEST_BLOCK=1`
- `git diff --check`

All fork-cloud matrix jobs passed.

Fork cloud validation for exact commit `860191a0397753e8790352be273f5a2b00eb523a`:

- [SwiftPM](https://github.com/rksharma-owg/gtm-session-fetcher/actions/runs/35228834834)
- [CocoaPods](https://github.com/rksharma-owg/gtm-session-fetcher/actions/runs/35228837999)
- [Bazel](https://github.com/rksharma-owg/gtm-session-fetcher/actions/runs/35228840950)
- [TestApps](https://github.com/rksharma-owg/gtm-session-fetcher/actions/runs/35228844150)

Fork cloud validation for previous commit `fa76a4239b58009ce70d209cd42618fdfe043f73`:

- [SwiftPM](https://github.com/rksharma-owg/gtm-session-fetcher/actions/runs/35006535789)
- [CocoaPods](https://github.com/rksharma-owg/gtm-session-fetcher/actions/runs/35006537631)
- [Bazel](https://github.com/rksharma-owg/gtm-session-fetcher/actions/runs/35006539527)
- [TestApps](https://github.com/rksharma-owg/gtm-session-fetcher/actions/runs/35006541482)

Fixes #185


Sources/Core/GTMSessionUploadFetcher.m
```diff
@@ -977,7 +977,7 @@ - (void)beginFetchWithCompletionHandler:(GTMSessionFetcherCompletionHandler)hand
   [self setDelegateCallbackQueue:self.callbackQueue];
   self.completionHandler = handler;
 
-  if ([self isRestartedUpload]) {
+  if ([self isRestartedUpload] && self.testBlock == nil) {
```
thomasvl: On quick skim, since this isn't guarded with GTM_DISABLE_FETCHER_TEST_BLOCK, I think a developer could still set the testBlock and impact this conditional even with GTM_DISABLE_FETCHER_TEST_BLOCK defined.

thomasvl: Actually, I guess the code down at 1004 also doesn't check, so maybe this is ok.

Sources/Core/GTMSessionUploadFetcher.m
```diff
@@ -992,6 +992,10 @@ - (void)beginFetchWithCompletionHandler:(GTMSessionFetcherCompletionHandler)hand
   GTMSESSION_ASSERT_DEBUG(self.fetcherInFlight == nil, @"unexpected fetcher in flight: %@",
                           self.fetcherInFlight);
 
+  if (self.request == nil && _uploadLocationURL != nil) {
```
thomasvl: Why was this needed since it isn't limited to the testblock path?

rksharma-owg: Scoped this to the test-block path. The request is now synthesized only when `hasTestBlock` is true, and `hasTestBlock` is forced to `NO` when `GTM_DISABLE_FETCHER_TEST_BLOCK` is enabled. Updated in fa76a42.

Sources/Core/GTMSessionUploadFetcher.m
rksharma-owg: Thanks for flagging this. I made the effective `hasTestBlock` value compile-time aware: it is always `NO` when `GTM_DISABLE_FETCHER_TEST_BLOCK` is enabled, so the restarted-upload conditional does not read or respond to a disabled test block. Test-block propagation to chunk fetchers is also guarded. Both normal and `-DGTM_DISABLE_FETCHER_TEST_BLOCK=1` builds pass locally, and the exact-head plus current-main fork matrices pass all 52 jobs at `fa76a42`.

Sources/Core/GTMSessionUploadFetcher.m
rksharma-owg: Restarted fetchers created by `uploadFetcherWithLocation:` have no initial request, while the superclass needs one to invoke `testBlock`. Commit `fa76a42` now synthesizes that request only when `hasTestBlock && self.request == nil && _uploadLocationURL != nil`; ordinary restarted uploads continue through the offset-query path. The regression test passes in the 26-job exact-head matrix and the 26-job merge matrix against current `main`.

thomasvl: Lets reduce this to 1-2 lines and put it in a comment there to explain for anyone reading the code in the future.

rksharma-owg: Added a 2-line comment above the block in 860191a explaining why the request is synthesized for `testBlock` on restarted fetchers. Thanks!

thomasvl (approved):
Thank you!

rksharma-owg:
All 26 matrix jobs across the 4 validation suites have completed and passed cleanly in GitHub Cloud VMs for commit `860191a`:

- [SwiftPM (run 35228834834)](https://github.com/rksharma-owg/gtm-session-fetcher/actions/runs/35228834834)
- [CocoaPods (run 35228837999)](https://github.com/rksharma-owg/gtm-session-fetcher/actions/runs/35228837999)
- [Bazel (run 35228840950)](https://github.com/rksharma-owg/gtm-session-fetcher/actions/runs/35228840950)
- [TestApps (run 35228844150)](https://github.com/rksharma-owg/gtm-session-fetcher/actions/runs/35228844150)

All 29 upstream PR checks are also green.

PR #584, opened 2026-09-14
Add option to disable JSON reformatting in debug logs (#101)

### LINKED ISSUE #101: Don't Automatically Prettify JSON Data in Log
Prettifying JSON data for the log is nice, but there should at least be a way to turn it off since it can cause the data to change.

For example, you can post content that looks like this:
```
{"number":2.06}
```

but it will be logged as:
```
{
    "number" : 2.0600000000000001
}
```

This can cause problems if you have a bug on your server but you spend all your time trying to debug the client because you have a log "proving" that's where the problem is.

rksharma-owg:
### Add option to disable JSON reformatting in debug logs ([#101](https://github.com/google/gtm-session-fetcher/issues/101))

- **Issue**: Debug logging unconditionally reformats and pretty-prints JSON responses via `NSJSONSerialization dataWithJSONObject:... options:NSJSONWritingPrettyPrinted`. This roundtrips JSON through Foundation objects, altering exact floating-point numbers (e.g. `2.06` becoming `2.0600000000000001`) and altering wire-level whitespace formatting.
- **Fix**:
  - Added `+setReformatLoggedJSONEnabled:` and `+isReformatLoggedJSONEnabled` to `GTMSessionFetcher (GTMSessionFetcherLogging)`, defaulting to `YES` for backward compatibility.
  - Added corresponding stubs in `GTMSessionFetcher` for when logging is stripped via `STRIP_GTM_FETCH_LOGGING`.
  - When reformatting is disabled (`NO`), the raw response string is returned verbatim if no token scrubbing was needed (preserving exact numbers, precision, and whitespace).
  - If sensitive tokens (`refresh_token`, `access_token`) are present and reformatting is disabled, the entire payload in the log is redacted (`_snip_`) rather than re-serialized, guaranteeing sensitive tokens are never logged while avoiding re-serialization.
- **Unit Test**: Added `testReformatLoggedJSON` in `UnitTests/GTMSessionFetcherUtilityTest.m`, with exact expected output for pretty-printed JSON, raw number preservation (`2.06`), redaction while reformatting is enabled, and full-payload redaction while reformatting is disabled.
- **Cloud validation**: The exact commit `d761683` and merge commit `28d68ee` against upstream `main` at `8b864f3` passed the 26-job macOS 26 matrix: SwiftPM/Xcode (12), CocoaPods (10), Bazel (2), and TestApps (2).

Fixes #101


Sources/Core/GTMSessionFetcherLogging.m
```diff
@@ -276,21 +289,40 @@ - (NSString *)formattedStringFromData:(NSData *)inputData
...
+      if ([[self class] isPrettifyLoggedJSONEnabled]) {
+        NSData *data = [NSJSONSerialization dataWithJSONObject:obj
+                                                       options:NSJSONWritingPrettyPrinted
+                                                         error:NULL];
+        if (data) {
+          NSString *jsonStr = [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding];
+          return jsonStr;
+        }
+      } else {
+        if (didRedact) {
```
thomasvl, line 313: I'm not sure this change as is worth while because of this edge case. If redaction is needed, then the original comment about things not being logged as they were on the wire would still apply.

I guess one way to try and do this would be if it was redacted, then just redact the whole log when they didn't want reformatting.  If we do that, then the option should likely be about reformatting and not the pretty print specifically and the api to control it should get a note that if the payload has something that needs redacting, then without reformatting, the whole payload is redacted in the log.

rksharma-owg: That makes complete sense! Redacting the whole payload when reformatting is disabled completely resolves the edge case and guarantees we never fall back to re-serializing or altering the wire data.

Updated in commit `03da9ce`:
- **Renamed API to reformatting**: Renamed methods to `setReformatLoggedJSONEnabled:` and `isReformatLoggedJSONEnabled`.
- **Full redaction on sensitive tokens**: When reformatting is disabled and the payload contains sensitive tokens (`access_token` or `refresh_token`), the entire payload in the log is redacted (`_snip_`) rather than re-serialized via `NSJSONSerialization`.
- **Documented edge case**: Added documentation to the header clarifying that if the payload has tokens requiring redaction, the entire payload is redacted in the log when reformatting is disabled.
- **Updated unit test**: `testReformatLoggedJSON` verifies that sensitive payloads have the entire payload redacted (`_snip_`) when reformatting is disabled, while non-sensitive payloads preserve exact wire formatting.

UnitTests/GTMSessionFetcherUtilityTest.m
```diff
@@ -29,11 +29,60 @@ @interface GTMSessionFetcher (GTMSessionFetcherLoggingInternal)
...
+  NSString *jsonInputString = @"{\"number\":2.06,\"message\":\"hello\"}";
+  NSData *jsonData = [jsonInputString dataUsingEncoding:NSUTF8StringEncoding];
+
+  // Default behavior (reformatting enabled) produces multi-line indented JSON.
+  NSDictionary *outJSON = nil;
+  NSString *prettyOutput = [fetcher formattedStringFromData:jsonData
+                                                contentType:@"application/json"
+                                                       JSON:&outJSON];
+  XCTAssertNotNil(prettyOutput);
+  XCTAssertTrue([prettyOutput containsString:@"\n"]);
```
thomasvl: With reformatting you should be able to hardcode what to test against to ensure it is exactly as expected.

UnitTests/GTMSessionFetcherUtilityTest.m
```diff
@@ -29,11 +29,60 @@ @interface GTMSessionFetcher (GTMSessionFetcherLoggingInternal)
...
+
+  NSDictionary *rawOutJSON = nil;
+  NSString *rawOutput = [fetcher formattedStringFromData:jsonData
+                                             contentType:@"application/json"
+                                                    JSON:&rawOutJSON];
+  // Verify it returns the raw string without roundtripping numbers (preserving 2.06) or adding indentation.
+  XCTAssertEqualObjects(rawOutput, jsonInputString);
+  XCTAssertEqualObjects(rawOutJSON[@"message"], @"hello");
+
+  // Verify that when reformatting is disabled, a payload containing sensitive OAuth tokens has the entire
```
thomasvl: Might as well make sure we have tests for redaction with reformatting.

UnitTests/GTMSessionFetcherUtilityTest.m
rksharma-owg: Updated in d761683. The enabled path now compares the complete pretty-printed string, including the `2.06` roundtrip to `2.0600000000000001`, instead of checking only for a newline. The [exact-head matrix](https://github.com/rksharma-owg/gtm-session-fetcher/actions?query=branch%3Acodex%2Fpreflight-pr584-d761683) and [current-main merge matrix](https://github.com/rksharma-owg/gtm-session-fetcher/actions?query=branch%3Acodex%2Fpreflight-pr584-merge8b864-d761683) both passed all 26 jobs.

UnitTests/GTMSessionFetcherUtilityTest.m
rksharma-owg: Added in d761683. With reformatting enabled, the test now passes an `access_token` payload and asserts the exact pretty-printed redacted JSON plus the redacted parsed value. The [exact-head matrix](https://github.com/rksharma-owg/gtm-session-fetcher/actions?query=branch%3Acodex%2Fpreflight-pr584-d761683) and [current-main merge matrix](https://github.com/rksharma-owg/gtm-session-fetcher/actions?query=branch%3Acodex%2Fpreflight-pr584-merge8b864-d761683) both passed all 26 jobs.

thomasvl (approved):
Thank you!

PR #581, opened 2026-09-14
Update downloadedLength during destinationFileURL download tasks (#88)

rksharma-owg:
### Update `downloadedLength` during download tasks ([#88](https://github.com/google/gtm-session-fetcher/issues/88))

- **Issue**: When downloading directly to a destination file (`destinationFileURL`), `_downloadedLength` remained `0` during the active download and was only updated upon download completion in `didFinishDownloadingToURL:`.
- **Fix**:
  - Update `_downloadedLength = totalBytesWritten;` inside `-[GTMSessionFetcher URLSession:downloadTask:didWriteData:totalBytesWritten:totalBytesExpectedToWrite:]`.
  - Update `_downloadedLength = fileOffset;` inside `-[GTMSessionFetcher URLSession:downloadTask:didResumeAtOffset:expectedTotalBytes:]`.
- **Unit Test**: Added assertions in `testFetchToFile` in `UnitTests/GTMSessionFetcherFetchingTest.m` to ensure `downloadedLength` tracks bytes written throughout the download and equals the total length on completion.

Fixes #88

PR #585, opened 2026-09-14
Guard destinationFileURL mutation during active fetch and track progress in simulated downloads (#60)

### LINKED ISSUE #60: GTMSessionFetcher seems to be accumulating data, even though a a download path is specified
We are using the GTMSessionFetcher (1.1.6) in order to download files from Google Drive.  We are having memory issues with large files.

Previously we were using GTMFetcher, but we updated our libraries.  In using the session fetcher, we are specifying a DownloadProgressBlock, DestinationFileURL and completion handler, like the following:

fetcher = [weakSelf.driveService.fetcherService fetcherWithURLString:url];

[fetcher setDownloadProgressBlock:downloadProgressHandler];
NSURL *myURL = [NSURL fileURLWithPath:fullpath];
[fetcher setDestinationFileURL:myURL];
[fetcher beginFetchWithCompletionHandler:completionHandler];

Everything downloads okay, but the memory increases and, for large files, runs out of memory.  

Previously, we had the following, which did not have the memory problem:
fetcher = [weakSelf.driveService.fetcherService fetcherWithURLString:url];

[fetcher setReceivedDataBlock:receivedData]; // setReceivedDataBlock is no longer available in the updated library
[fetcher setDownloadPath:fullpath]; // setDownloadPath is no longer available in the updated library
[fetcher beginFetchWithCompletionHandler:completionHandler];

Is there some additional configuration needed, so that data is not accumulated?  I tried [fetcher setAccumlatedDataBlock:nil], but that seemed to have no effect.

Thank you!
-Kevin

rksharma-owg:
### Guard `destinationFileURL` mutation during active fetch and track progress in simulated downloads ([#60](https://github.com/google/gtm-session-fetcher/issues/60))

- **Issue**:
  1. Modifying `destinationFileURL` after `beginFetch` has been invoked can cause unexpected transitions or data loss.
  2. In simulated downloads with a destination file via `testBlock`, `_downloadedLength` was not updated during simulated chunk callbacks or write completion.
- **Fix**:
  - Added debug assertion in `-[GTMSessionFetcher setDestinationFileURL:]` checking that `![self isFetchingUnsynchronized]`.
  - In `-[GTMSessionFetcher simulateDataCallbacksForTestBlockWithBodyData:bodyData:]`, updated `_downloadedLength = totalBytesDownloaded;` inside simulated `downloadProgressBlock` and `_downloadedLength = (int64_t)responseData.length;` on simulated write completion.
- **Unit Test**: Added `testDestinationFileURLCannotBeChangedAfterBeginFetch` in `UnitTests/GTMSessionFetcherFetchingTest.m` and added `downloadedLength` assertions to `testFetcherTestBlockSimulateDataCallbacks`.

Fixes #60


thomasvl:
I'm not sure I follow what this PR has to do with #60. That issue is about memory growth of the process, but this PR seems more ensuring the _downloadedLength is updated; which makes me wonder if some of it belonged in #581 also?

rksharma-owg:
This does not address the process-memory growth reported in #60. The destination mutation guard is a separate behavior, and the simulated `downloadedLength` updates overlap the area covered by #581 without demonstrating the reported leak. I’m closing this rather than relabeling or broadening an unsupported fix.

